PR #11, opened 2016-12-09
Make serve.sh executable, mention it in README

zachmargolis:
Realized there was a `serve.sh` script here, but it wasn't executable. Also it wasn't mentioned in the README.

Maybe we don't need it since it's just a light wrapper for `jekyll serve` anyways, and we could remove it entirely?


blacktm:
Yea, I added that to save keystrokes typing `jekyll serve --baseurl ""` each time, and setting the `baseurl` means the site is simply served at the root and you don't have to append the `identity-dev-docs/` locally.

zachmargolis:
So @blacktm is it OK to make `serve.sh` the main way to develop?

blacktm:
Sure, I'll merge.

PR #12, opened 2016-12-19
Put SP config attributes in a table

blacktm:


blacktm:
Cool, thanks @jessieay.
@amoose Ok if we merge this? (Feel free to do the honors if so.)

jessieay:
Done! On end user we are mostly self-merging so I wasn't sure if that was a style you were used to...

blacktm:
@jessieay Got it, that sounds good.

PR #14, opened 2017-02-03
OpenID Connect Documentation

zachmargolis:
Documents the current state of OpenID connect (PKCE + no client secret)

I didn't link this from anywhere yet. Also at some point I'd like to reorganize the main index page so that we can link to SAML and OpenID Connect as peers but I don't think we're quite ready for that.


_pages/open_id_connect.md
```diff
@@ -0,0 +1,246 @@
...
+     "street_address": "123 Main St. Apt 123",
+     "locality": "Washington",
+     "region": "DC",
+     "postal_code": "20001"
+   }
+   ```
+
+ * **birthdate** *requires the `profile` scope and an LOA 3 account*
+
+   Birthdate, formatted as ISO8601‑2004 `YYYY-MM-DD`.
```
pkarman: I've never seen this syntax `ISO8601‑2004` -- did you mean `ISO 8601:2004` ?

zachmargolis: Yes, that is what I meant, thanks!

PR #15, opened 2017-02-06
Move OpenID Connect docs URL

zachmargolis:
The top-level directory uses dashes (`identity-dev-docs`) so I think the other pages should use dashes too

### DIFF
--- a/_pages/openid_connect.md
+++ b/_pages/openid_connect.md
@@ -1,5 +1,6 @@
 ---
 title: OpenID Connect
+permalink: /openid-connect
 ---
 
 <div class="usa-alert usa-alert-warning">


_pages/openid_connect.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: OpenID Connect
+permalink: /openid-connect
```
blacktm: Actually, the way this Jekyll site is configured, the path is automatically equal to the file name, so you can remove `permalink` entirely (like the other pages).

blacktm: Also, the way [Federalist](https://federalist.18f.gov) works (our homegrown pages server), all internal links need a trailing `/`, so if you _do_ use a permalink, need to add that. Just FYI.

PR #17, opened 2017-02-09
Commit Gemfile.lock

zachmargolis:
**Why**:
Federalist will use the jekyll version from the Gemfile.lock

--

pro tip frmo @wslack and @jmhooper


wslack:
defer to @jmhooper

blacktm:
Has this been an issue? Haven't noticed any problems publishing yet. I don't typically track specific gems or lock down the Ruby version on Jekyll sites.

zachmargolis:
In Slack, @jmhooper said that we could check it in to make sure we're deploying with the same version of Jekyll we're developing against, so it seemed like a good idea?

monfresh:
The general consensus in the Ruby community is that `Gemfile.lock` should always be checked in in any Ruby app that has gem dependencies. The only exception is if the repo is a gem. http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

blacktm:
I mean, yes, but this isn't a Ruby app, just content — doesn't even need a Gemfile really. If the goal is to ensure that local rendering matches what will be rendered by Federalist, then it'd be nice to see what their gem dependencies are and conform to them ([GitHub pages publishes theirs](https://pages.github.com/versions/)).

zachmargolis:
This has been sitting for a bit, and it looks like we came to a consensus that checking in `Gemfile.lock` is OK, so I'm going to merge this. Let me know if we need to revert.

PR #18, opened 2017-02-09
Add OIDC examples for both PKCE and private_key_jwt

zachmargolis:
Also add a fancy picker!

picker

### DIFF
--- a/_pages/openid-connect.md
+++ b/_pages/openid-connect.md
@@ -12,15 +12,18 @@ permalink: /openid-connect/
 
 # OpenID Connect 1.0 Developer Guide
 
-login.gov supports OpenID Connect 1.0, an extension of Oauth 2.0, conforming to the [iGov Profile][igov-profile].
+login.gov supports [OpenID Connect 1.0][openid-connect], an extension of Oauth 2.0, conforming to the [iGov Profile][igov-profile].
 
+[openid-connect]: http://openid.net/specs/openid-connect-core-1_0.html
 [igov-profile]: http://openid.net/wg/igov/
 
 In this guide:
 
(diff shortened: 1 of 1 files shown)


_pages/openid-connect.md
```diff
@@ -271,3 +340,32 @@ login.gov supports some of the [standard claims from OpenID Connect 1.0][standar
 
  * **sub**
    Unique ID for this user. This ID is unique per client.
+
+<script type="text/javascript">
```
pkarman: time for jquery or its ilk?

_pages/openid-connect.md
```diff
@@ -52,18 +69,34 @@ Users need to authorize OpenID Connect 1.0 clients individually. To present the
 
 ### Authorization Request
 
-```
+View example as <button data-example="pkce">PKCE</button><button data-example="private_key_jwt">private_key_jwt</button>
```
pkarman: Do we need to define some acronyms? Or are they already defined elsewhere in this doc? "PKCE" and "JWT" etc.

zachmargolis: The first few times I use JWT, I link to the JWT spec?

_pages/openid-connect.md
zachmargolis: This is simple enough, I think it's easier than adding a dependency on jQuery, unless you feel strongly? `document.querySelectorAll`, a new standard, is inspired by jQuery?

pkarman: no strong feelings. Just seeing that much raw JS made me wonder.

_pages/openid-connect.md
zachmargolis: I added a link to the PKCE spec as well

PR #20, opened 2017-02-22
Remove glossary

zachmargolis:
**Why**:
It's out of date compared to the page in the identity-idp wiki
(https://github.com/18F/identity-idp/wiki/Glossary)

---

We also don't link to this from anywhere as far as I can tell


blacktm:
Dang it, bumped the close button. Yes, that was left over content, so 👍 to remove.

PR #22, opened 2017-02-23
Add SAML guide with details per-endpoint

zachmargolis:
**Why**: Create a reference similar to the OpenID Connect one

---

This is a prototype of another way I was thinking about organizing our docs. For OpenID Connect, I went with a guide that walks through things per endpoint: https://pages.18f.gov/identity-dev-docs/openid-connect/

@jessieay, would love your feedback on if a page like this makes any sense, since you've been through this yourself?

@amoose would love a check on the content, since my understanding on SAML is still a little fuzzy

### DIFF
--- a/_pages/saml.md
+++ b/_pages/saml.md
@@ -1,0 +1,263 @@
+---
+title: SAML
+permalink: /saml/
+---
+
+# SAML Developer Guide
+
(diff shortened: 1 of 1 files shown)


zachmargolis:
@jessieay yeah, I'll look into incorporating the "configuring your SP" page. I have a separate PR that updates the "registering your SP" page here: https://github.com/18F/identity-dev-docs/pull/23

zachmargolis:
Ok I embedded some of the information from "configuring your SP" -- I'm going to merge this, it's not linked to from anywhere, so the old pages will still be up. I'm working on another branch that will unify the old navigation

PR #26, opened 2017-03-03
Add additional OpenID Connect scopes, profile:name and profile:birthdate

zachmargolis:
see: https://github.com/18F/identity-idp/pull/1164

### DIFF
--- a/_pages/openid-connect.md
+++ b/_pages/openid-connect.md
@@ -137,6 +137,8 @@ https://idp.int.login.gov/openid_connect/authorize?
    - `address`
    - `email`
    - `phone`
+   - `profile:birthdate`
+   - `profile:name`
    - `profile`
    - `social_security_number`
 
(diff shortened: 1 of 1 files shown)


_pages/openid-connect.md
```diff
@@ -137,6 +137,8 @@ https://idp.int.login.gov/openid_connect/authorize?
    - `address`
    - `email`
    - `phone`
+   - `profile:birthdate`
```
pkarman: IIRC we call it `dob` on the SAML side. It would be nice for consistency in all docs and APIs, and I don't care which one we pick. Fine with me too to merge this as-is and change later, unless you feel like This Is The Moment to clarify.

zachmargolis: This was me matching the openid connect `birthdate` attribute, so it made sense to me to ask for `birthdate` and get `birthdate` back?

pkarman: we should probably rename SAML to use `birthdate` then. I will open ticket.

PR #27, opened 2017-03-03
Update attributes table to show SAML and OIDC names

zachmargolis:
screen shot 2017-03-03 at 2 10 00 pm


pkarman:
+1 super helpful. Esp since we can now see *all* the differences in naming.

PR #33, opened 2017-04-04
Remove work-in-progress banner

zachmargolis:
I think this is stable enough to read/use now.

openid_connect


blacktm:
😆 That's a very clear screen shot, much appreciated.

PR #39, opened 2017-05-31
Add anchors to OIDC attributes

zachmargolis:
@blacktm curious what you think of this, if it would be easy to link people to specific anchors for each parameter for OIDC? Just an idea, don't need to merge it if it wouldn't be useful.


blacktm:
Sure, can be useful. I send a lot of links to agencies.

PR #40, opened 2017-06-08
Update titles to be more consistent

zachmargolis:
Just some cleanup

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,5 @@
 # Site settings
-title: Login.gov Developer Docs
+title: login.gov
 description: Welcome to the Login.gov developer documentation!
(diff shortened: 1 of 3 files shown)


hursey013:
The end result is the same, but if you don't want to have to prepend each title with `login.gov` we could just update the title in `_config.yml` to `login.gov` and make the change globally here: https://github.com/18F/identity-dev-docs/blob/18f-pages/_layouts/main.html#L9

`<title>{% if page.title %}{{ site.title }} | {{ page.title }}{% else %}{{ site.title }}{% endif %}</title>`

zachmargolis:
@hursey013 great idea! Updated

PR #45, opened 2017-06-28
Deploy latest changes

hursey013:


zachmargolis:
I switched this repo to be served from `master` not `18f-pages`, so these changes are already in prod 😀

hursey013:
Oh nice! I guess I wasn't patient enough w/ Federalist - I didn't see the changes at first, but just reloaded and there they are! Thanks

PR #48, opened 2017-07-28
Update registration instructions

zachmargolis:
**Why**: Make it clearer how to get in touch


zachmargolis:
@andrewhughey I figured we would email them instructions for how to log in? Would love to link directly to it, but since it's behind basic auth, probably not useful

PR #50, opened 2017-07-28
Update self-signed cert instructions

zachmargolis:
**Why**: Customer feedback

--

cc @lauraGgit

### DIFF
--- a/_pages/certs.md
+++ b/_pages/certs.md
@@ -9,10 +9,10 @@ First, you'll need to create an `openssl.conf` file, for example:
 ```
 [ req ]
 default_bits           = 2048
-distinguished_name     = REQ_DISTINGUISHED_NAME
+distinguished_name     = req_distinguished_name
 prompt                 = no
 
-[ REQ_DISTINGUISHED_NAME ]
+[ req_distinguished_name ]
 commonName             = YOUR_DOMAIN
 organizationName       = GSA
 organizationalUnitName = 18f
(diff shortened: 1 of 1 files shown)


_pages/certs.md
```diff
@@ -9,11 +9,11 @@ First, you'll need to create an `openssl.conf` file, for example:
 ```
 [ req ]
 default_bits           = 2048
-distinguished_name     = req_distinguished_name
+distinguished_name     = REQ_DISTINGUISHED_NAME
```
lauraGgit: Actually looks like this does not generally change.

_pages/certs.md
```diff
@@ -9,11 +9,11 @@ First, you'll need to create an `openssl.conf` file, for example:
...
 prompt                 = no
 
-[ req_distinguished_name ]
-commonName             = idp-sandbox.login.gov
+[ REQ_DISTINGUISHED_NAME ]
```
lauraGgit: Similarly, looks like it should stay the same.

_pages/certs.md
```diff
@@ -22,7 +22,10 @@ countryName            = US
 emailAddress           = 18f@gsa.gov
 ```
 
-You'll need to replace `req_distinguished_name` with your agency's details.
+You'll need to replace a few parts:
```
lauraGgit: Might suggest the following:
```
You'll need to update the information listed under the `[ req_distinguished_name ]` section in the `.conf` to match your details. 

The `commonName` will need to reflect the domain that your SP will be initiating the SAML request from.
```

zachmargolis: Ok thanks! Updated in ca95f5722e4af172c904804ab975d46868fea928, PTAL!

lauraGgit:
LGTM! 🚀

PR #52, opened 2017-09-05
Add link to ASP/.NET OIDC example

zachmargolis:


blacktm:
Awesome, was meaning to add that (they style it [ASP.NET](https://www.asp.net), but no biggie). Thanks!

zachmargolis:
whoops, will fix!

PR #57, opened 2018-01-31
Edit Button

s2t2:
Additions:

  + Button at the bottom of each page which links back to the respective GitHub edit page link.
  + Includes Font Awesome icons (fontawesome-free-5.0.6)

Screenshot:

screen shot 2018-01-31 at 4 38 01 pm


blacktm:
That's a lot of JS to add. Maybe we can just find a small SVG, or a ✏️ emoji.

s2t2:
OK I removed the JS file and instead took an approach to use only a subset of the Font Awesome icons. The nice thing about having an icon set available is it will prevent developers in the future from needing to search for one-off icons which may or may not match the rest.

blacktm:
Cool, just made a couple small fixes, namely loading the resources from the repo itself (we try to limit non-gov dependencies) and use of the JS/SVG version for enhanced accessibility.

PR #59, opened 2018-02-08
Restore key-generation instructions

s2t2:
Additions:

  + Includes key-pair generation instructions adapted from previous `certs` page. I found these instructions helpful when creating the `expressjs` client application, and also when getting an agency partner started with development.

Revisions:

  + Formally recognizes the well-known configuration as an endpoint. This is helpful when understanding the login flow.

Considerations:

  + Includes JWT-specific language.


blacktm:
I like the configuration guidance! Since the certs/public key also applies to SAML, maybe we should restore the page old page entirely. Doesn't need its own top-level nav, but maybe a link in both OIDC and SAML guides that say, "to generate a public/private key for testing your integration, do this" or something.

orenyk:
Closing in lieu of #163

PR #61, opened 2018-03-15
Use updated, accurate government banner content

maya:
The government banner text is using older, inaccurate content and this fixes that. 

Current banner: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/develop/components/detail/banner.html

See: https://github.com/uswds/uswds/issues/2093


blacktm:
Awesome, thanks @maya! Also updated our dashboard, https://github.com/18F/identity-dashboard/pull/165

PR #67, opened 2019-02-06
Use identity-style-guide.

adunkman:
This PR switches to use `identity-style-guide@1.0.1` for static assets. It’s a pretty standard layout switch — but I’ll drop a few inline comments below. 

I’ve noticed a few things that I’d like to change about the styles, but I’m not sure any are bad enough to block merging — but we can definitely have that chat, should you spot anything that’s a deal-breaker. (It should be quite straightforward to upgrade to the next version of `identity-style-guide`, so it shouldn’t be seen as another large effort). 

~**One potentially blocking issue** — is this deployed through Federalist? If so, we may need to tweak the federalist build script to run `npm install` — but it may also pick it up automatically, as `identity-style-guide` itself did. If it is deployed with Federalist, I’d like to enable building PRs to check this.~ (It is deployed on Federalist, and the webhooks are now working, and everything looks great!)

Here’s a few pages side-by-side, to aid in review: 

| Before | After |
|-------|-------|
| developers login gov_ | localhost_4000_ |
| developers login gov_faq_ | localhost_4000_faq_ |
| developers login gov_oidc_ | localhost_4000_oidc_ |


.circleci/config.yml
```diff
@@ -5,26 +5,28 @@ jobs:
   build:
     docker:
       # Specify the Ruby version you desire here
-      - image: circleci/ruby:2.5.0
+      - image: circleci/ruby:2.5.3-node
```
adunkman: Switching the base image to Ruby 2.5.3 includes a more up-to-date version of `npm`, which enables `npm ci`. Can also install it manually during the build process, if we’d rather stick with Ruby 2.5.0.

_config.yml
```diff
@@ -9,6 +9,9 @@ plugins:
 whitelist:
   - jekyll-redirect-from
 
+copy_to_destination:
+  - node_modules/identity-style-guide/dist/assets
+
```
adunkman: Unfortunately, for [Jekyll configuration](https://jekyllrb.com/docs/configuration/options/), the `exclude` filter [takes precedence over](https://github.com/jekyll/jekyll/issues/4791) the `include` filter. I’d rather specifically include this one folder, rather than including all of `node_modules` (excluded by default).

Copying a bunch of static files can seriously slow down the build process, so I think this is the right way to go (the custom plugin) — but let me know if you think it’s something worth revisiting, or if you know of a better way!

_pages/faq.md
```diff
@@ -2,126 +2,175 @@
 title: Frequently asked questions
 ---
 
-# Frequently asked questions
+<div class="usa-accordion usa-accordion-bordered" aria-multiselectable="true">
```
adunkman: This page previously wrapped each question in its own accordion — this switches to a single accordion with `aria-multiselectable="true"`, which keeps the same behavior (more than one question can be expanded at the same time).

I’m not sure this is a great use of accordions in general, but that’d be a separate PR!

stevegsa:
Great work!

PR #88, opened 2019-10-28
Remove Java and Python example SPs

blacktm:
These are old and unmaintained, removing references from them in our developer documentation


amoose:
Yes! Please add a note that these are **not** reference implementations and are non-production examples, only. We should also add a note to each sample's README

blacktm:
@amoose Ah, good points!

PR #102, opened 2020-02-12
LG-2606: Upgrade developer docs to newer identity-style-guide

zachmargolis:
This is a first pass, I was able to match most of the styles. It looks like a lot of the weights and spacing changed, and I think that's intentional? It's definitely easier to tell by spinning it up locally & tabbing between before and after

BEFORE
before-home
before-oidc

AFTER

after-home
after-oidc

PR #105, opened 2020-02-13
Dockerize the app

ctro:
- Automate uswds install via npm


ctro:
@amoose Snyk has an issue with something here, but when I try to follow the `Details` I see

```
Unable to display this organization
The organization does not exist, or you do not have permissions to access it.
```

Any ideas?

package.json
```diff
@@ -11,6 +14,8 @@
   },
   "homepage": "https://github.com/18F/identity-dev-docs#readme",
   "dependencies": {
-    "identity-style-guide": "^1.1.0"
+    "identity-style-guide": "^1.1.0",
+    "uswds": "^2.0.2",
```
zachmargolis: Do we use uswds directly? I thought that came with as part of identity-style-guide?

ctro: Yeah, I'm not sure! Does @nickttng know the answer?

AFAIK adding `uswds` here just puts it in node_modules.  The new npm task/script I added copies the `uswds` source into the same place that the README specifies.

🤔 @nickttng Should I be copying files from `node_modules/identity_style_guide`?

zachmargolis: So for my PR yesterday to update `identity-style-guide`, I only had to change that one node dependency to get the extra styles: https://github.com/18F/identity-dev-docs/pull/102

I'm gonna check out your branch and see how things build if I take out that dep

README.md
```diff
@@ -2,23 +2,53 @@
...
+docker-compose run web bash
 ```
 
-## Updating the template
+### Locally
 
-This site uses the [U.S. Web Design Standards](https://standards.usa.gov). To update them:
+Run the server locally: 
 
-1. Download the [latest release](https://standards.usa.gov/download)
```
ctro: I didn't have any styles in the local site until I grabbed uswds via npm and copied stuff over per these instructions.

zachmargolis: 🤷‍♂ I took out the `RUN npm run build-uswds` and have the styles somehow, but I may have a dirty environment....going to do a git clean and try again

zachmargolis: Ok! So with a clean environment, the styles didn't load automatically.

But then I did this:
```
docker-compose exec web bash
```
and ran `make build` inside the container, and things worked. So maybe the default `up` is missing something that `make build` or `make run` do?

package.json
zachmargolis: https://github.com/18F/identity-dev-docs/pull/106

README.md
ctro: OK, that makes sense. Checking out that other PR and will run it all in a fresh env

package.json
ctro: That PR totally works for me.  I'm don't 100% get _how_ yet, but it's def simpler! Thanks Zach!

nickttng: @ctro We should be using `"identity-style-guide": "^2.1.6"` which currently uses `uswds@2.0.3`, so there's no need to list USWDS directly in the dependencies. Hope this clears up.

ctro: Yes, this helps!  But I'm still a little 🤔 about how identity-style-guide and USWDS work together.  I'm going to put some time on our calendars so you can teach me about it ❤️

PR #106, opened 2020-02-13
Simplify package.json dependencies, re-use Makefile in Docker

zachmargolis:
Give this a shot?


Makefile
```diff
@@ -1,5 +1,5 @@
 run: install-dependencies
-	bundle exec jekyll serve
+	bundle exec jekyll serve --host 0.0.0.0
```
zachmargolis: Maybe I should have carried over the `--incremental`?

PR #108, opened 2020-02-14
LG-2567 IALMAX docs

stevegsa:


Danielle-Lee:
On the openID Connect page, do we need the definition of LOA? If so, we should also define IAL. 

Current: "The Authentication Context Class Reference values used to specify the LOA (level of assurance) of an account, either LOA1, LOA3, or IAL0 (also known as IALMAX)" 

Changes to "The Authentication Context Class Reference values used to specify the LOA (level of assurance) of an account as either LOA1, LOA3 or IAL0 (Identity Level of Assurance 0), also known as IALMAX."  

No other changes.

porta-antiporta:
@stevegsa @Danielle-Lee @jmhooper I noticed we don't have ACR value information for IAL1/IAL2 in prod.  Is that intentional?  I thought we had made the changes to support asserting IAL values in the acr field? (We had to roll back, but have since fixed it).  cc: @jennywagg

_pages/oidc.md
```diff
@@ -54,9 +54,10 @@ Consistent with the specification, login.gov provides a JSON endpoint for OpenID
...
 
 * **acr_values**
-  The Authentication Context Class Reference values used to specify the LOA (level of assurance) of an account, either LOA1 or LOA3. This and the `scope` determine which [user attributes]({{ site.baseurl }}/attributes/) will be available in the [user info response](#user-info-response). The possible parameter values are:
+  The Authentication Context Class Reference values used to specify the LOA (level of assurance) of an account, either LOA1, LOA3, or IAL0 (also known as IALMAX). If IALMAX is chosen the request will return IAL1 data if a user is not verified or IAL2 data if a user is verified. This and the `scope` determine which [user attributes]({{ site.baseurl }}/attributes/) will be available in the [user info response](#user-info-response). The possible parameter values are:
     - `http://idmanagement.gov/ns/assurance/loa/1`
     - `http://idmanagement.gov/ns/assurance/loa/3`
+    - `http://idmanagement.gov/ns/assurance/ial/0` (also known as IALMAX)
```
zachmargolis: +1 to what I think @porta-antiporta was asking, the code also allows `/ial/1` and `ial/2`, should we list those here?

porta-antiporta:
Found it it was [LG-2410 ](https://cm-jira.usa.gov/browse/LG-2410).  I recall we had made changes to include these values.  There was a PR #89 for dev docs as well, which seems to have been removed.

_pages/oidc.md
porta-antiporta: Old PR that had added it #89

porta-antiporta:
If we restore #89, I'd keep IAL0 to be associated to IAL1/IAL2 acr values (as we're trying to steer folks away from using LOA.  We only support LOA for backwards compatibility of legacy integrations.

_pages/oidc.md
stevegsa: weird.  so it somehow got removed?

porta-antiporta: We removed it when we had to roll back the change (the initial implementation to support IAL wasn't backwards compatible so it broke RRB). I suspect we may have omitted restoring it when we fixed the issue.

Would you be able to merge the changes in #89 to this branch? (Is that allowed in our SCM scheme?)

stevegsa: sure

PR #114, opened 2020-03-02
Document verified_at Attribute

zachmargolis:
Document changes added by

https://github.com/18F/identity-idp/pull/3602


_pages/attributes.md
```diff
@@ -6,13 +6,158 @@ lead: >
 
 Here are the possible attributes that can be requested at a given IAL. This table contains the available user attributes, the IAL they are associated with, and how they can be accessed in OpenID Connect and SAML.
 
-| Attribute | IAL1 | IAL2 | OpenID Connect | SAML |
```
jmhooper: No love for markdown tables here

jmhooper:
Hmm, it looks like Federalist blew up on the build. I logged into federalist and it looks like the build doesn't have any logs associated with it 🤔

_pages/attributes.md
jmhooper: It looks like the `verified_at` attribute did not actually get added to this table. Was that intentional?

jmhooper:
I queued up another build. Looks like that one was successful.

_pages/attributes.md
zachmargolis: Well I do like Markdown tables a lot but this was getting tough to read and edit.

And yes I forgot to add the attribute after I converted it :see_no_evil:

zachmargolis: Fixed in 491c75a50f869fea1ce00d6bfefc25639d837dd6

PR #127, opened 2020-05-14
Adding x509 attributes

ethanheppner:


ethanheppner:
@achapm made the edits-- lmk if anything else is needed!

ethanheppner:
@achapm are you good to approve and merge this?

_pages/attributes.md
```diff
@@ -177,6 +177,64 @@ Requires the `profile` or `profile:verified_at` scope.
 </td>
 <td markdown="1">
 `verified_at` (string, ISO8601 format)
+</td>
+    </tr>
+    -----------------------------
```
zachmargolis, line 182: Should we remove this line? I'm not seeing it rendered

zachmargolis:
On the OIDC page in the [User Info Response](https://developers.login.gov/oidc/#user-info-response) section we also document the attributes as they appear in OIDC, with some examples can we add these attributes there too?

_pages/attributes.md
```diff
@@ -177,6 +177,64 @@ Requires the `profile` or `profile:verified_at` scope.
...
+**x509 Presented** <br />
+</td>
+<td markdown="1">
+![checkmark][checkmark]
+</td>
+<td markdown="1">
+![checkmark][checkmark]
+</td>
+<td markdown="1">
+`x509_presented` (string)
```
zachmargolis, line 232: I'm having trouble tracking down an example value for this -- we're sure it's a string right?

_pages/attributes.md
ethanheppner: Yes, we can get rid of that-- I'd only put this in to keep my place and forgot to delete it afterwards.

_pages/attributes.md
ethanheppner: I'm not sure-- I just added the content @achapm suggested, but if something else seems more correct to you, feel free to put that in.

PR #132, opened 2020-07-01
Add id_token to auth response

achapm:
#WHY
Developers should see that id_token is included in the authorization response on OIDC docs
- Mention that id_token is signed using RS256

### DIFF
--- a/_pages/oidc.md
+++ b/_pages/oidc.md
@@ -264,7 +264,7 @@ The token response will be a JSON object containing the following:
   The number of seconds the access token will expire.
 
 * **id_token** (string)
-  A signed [JWT][jwt] that contains basic attributes about the user and it is signed using RS256. The public key used to verify this JWT is available from the [certificates](#certificates) endpoint.
+  A signed [JWT][jwt] that contains basic attributes about the user and it is signed using the **RS256** algorithm. The public key used to verify this JWT is available from the [certificates](#certificates) endpoint.
 
 Here's an example token response:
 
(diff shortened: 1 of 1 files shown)


_pages/oidc.md
```diff
@@ -195,6 +195,7 @@ In an **unsuccessful authorization**, the URI will contain the parameters `error
   - `invalid_request` — The authorization request was invalid. See the `error_description` parameter for more details.
 - **error_description** — A description of the error.
 - **state** — The `state` value originally provided by the client.
+- **id_token** - (See id_token in [token response](#token-response) for more detail)
```
zachmargolis, line 198: This is wrong. The redirect URI from the authorization point does not contain the ID token

_pages/oidc.md
```diff
@@ -264,7 +265,7 @@ The token response will be a JSON object containing the following:
   The number of seconds the access token will expire.
 
 * **id_token** (string)
-  A signed [JWT][jwt] that contains basic attributes about the user. The public key used to verify this JWT is available from the [certificates](#certificates) endpoint.
+  A signed [JWT][jwt] that contains basic attributes about the user and it is signed using RS256. The public key used to verify this JWT is available from the [certificates](#certificates) endpoint.
```
zachmargolis, line 268: Stylistically I would make this more terse and code-quote the algorithm

> ...contains basic attributes about the use, signed using `RS256`. The public key...

_pages/oidc.md
zachmargolis: Here's the code for the success redirect URI:
https://github.com/18F/identity-idp/blob/master/app/forms/openid_connect_authorize_form.rb#L87-L92

and the error redirect URI:
https://github.com/18F/identity-idp/blob/master/app/forms/openid_connect_authorize_form.rb#L194-L203

we never give back the `id_token` in the redirect URI because redirects are easily intercepted and we don't want to give away PII like that

PR #141, opened 2020-08-10
Remove border from navbar styling

zachmargolis:
Request from @juliaelman 

Before

Screen Shot 2020-08-10 at 10 48 23 AM

After

Screen Shot 2020-08-10 at 10 48 07 AM


_sass/design-overrides.scss
```diff
@@ -0,0 +1,3 @@
+.border-top-none {
+  border-top: none !important;
```
zachmargolis, line 2: so I'm not a huge fan of `!important` but the style this is competing with is very specific (`@media (min-width: 1024px) .usa-header--extended .usa-nav`) so even though this style was applied to the element, that one "won"

aduth: > so I'm not a huge fan of `!important` but the style this is competing with is very specific (`@media (min-width: 1024px) .usa-header--extended .usa-nav`) so even though this style was applied to the element, that one "won"

In terms of [CSS specificity score](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity), I'd tend to approach these sorts of overrides as a sort of "score to beat + 1", like as a modifier to the target being overridden.

Something like:

```css
.usa-header--extended .usa-nav.border-top-none {
  border-top: none;
}
```

Or inverted to make `.border-top-none` contextually-aware:

```scss
.border-top-none {
  .usa-header--extended .usa-nav& {
    border-top: none;
  }
}
```

The downside is it's very tied to the specific thing being overridden. But in most cases, that's kinda the point.

zachmargolis: I like that first approach! Making a PR now

PR #144, opened 2020-08-18
RISC Security Events Update Proposal

zachmargolis:
I figured I'd propose changes to our RISC implementation via documentation first. Will drop some specific things I have questions on as comments


_pages/security-events.md
```diff
@@ -20,25 +25,35 @@ sidenav:
...
+### Authorization Fraud Detected
+
+RPs should submit this event when they believe a user's credentials have been compromised, that somebody who is not the user
+was able to sign in to a user's account. login.gov may force the reset of user's password when we receive this event.
 
+The **event_type** for this is:
 
-### Account Credential Change Required
+```
+https://schemas.openid.net/secevent/risc/event-type/login-gov/authorization-fraud-detected
```
zachmargolis, line 46: So we're going to create new event types

1. the schemas.openid.net doesn't exist
2. I figured if I did a login.gov domain, people might expect it to exist, so I kept the old one
3. I added `/login-gov/` prefix to the event name, as a way to mark that it's ours?

Would love thoughts on if the actual domain of this should be login.gov

_pages/security-events.md
```diff
@@ -82,16 +97,25 @@ JWTs must be signed by the client application's private key using **RS256**, the
...
+        This is login.gov's issuer, the root URL for login.gov. In the agency integration environment, this is `https://idp.int.identitysandbox.gov`
 
-  * **subject_type**
-    Must be **iss_sub**, this indicates the **sub** is the subject provided by the original issuer (login.gov)
+      * **sub**
+        The UUID identifying the user. This is provided as the `sub` inside the `id_token` JWT in the [OpenID Token endpoint]({{site.baseurl}}/#token-response).
 
-  * **iss**
-    This is login.gov's issuer, the root URL for login.gov. In the agency integration environment, this is `https://idp.int.identitysandbox.gov`
+  * **event_at**
```
zachmargolis, line 116: I almost went with `occurred_at` but I kept spelling "occurred" wrong so I figured we could maybe prevent future typos this way. However, the name is not very descriptive, open to thoughts here

_pages/security-events.md
```diff
@@ -20,25 +25,35 @@ sidenav:
...
 
-### Account Credential Change Required
+```
+https://schemas.openid.net/secevent/risc/event-type/login-gov/authorization-fraud-detected
+```
 
-Submit this event to login.gov if the user should change their credential (usually, reset their password).
+### Identity Fraud Detected
+
+RPs should submit this event when they believe a verified identity may be fraudulent. login.gov may reset the user's profile and verified attributes data when we receive this event.
```
zachmargolis, line 51: I didn't know a better way to say "proofed as somebody else" but open to wording/phrasing here

_pages/security-events.md
mitchellhenke: Initial feeling is to have the URI be something that we _could_ make exist, even if it isn't necessary currently?

_pages/security-events.md
mitchellhenke: I think `event_at` is ok, a more verb-y option could be something like `discovered_at`?

zachmargolis: So my use case is, suppose the fraud happened 2 weeks ago, but we figured it out now and reported it now. I'm looking to capture the 2 weeks ago timestamp, but I am worried `discovered_at` would lead people to submit "now" since it's when they discovered the event?

_pages/security-events.md
zachmargolis: Ok so more like:

```
https://schemas.login.gov/secevent/risc/event-type/authorization-fraud-detected
```

_pages/security-events.md
mitchellhenke: Ohhh, right right. `event_at` I think could be too ambiguous. Maybe `happened_at` or `affected_at`?

_pages/security-events.md
mitchellhenke: Looks good to me, seems odd their schemas don't exist though.

_pages/security-events.md
zachmargolis: Yeah I like `happened_at` out of the two of those, but I'm warming back up to  `occurred_at`

mitchellhenke: I think either works 👍🏼

jmhooper: I also can't spell occurred, but I think it captures the meaning the best and am a fan.

_pages/security-events.md
jmhooper: My shot "RPs should submit this event when they suspect an account has been used to commit identity theft or related fraudulent activity."

zachmargolis: I like it! Thanks

PR #146, opened 2020-09-02
Update RISC pages

zachmargolis:
There's  a lot going on here! Some of these changes are still pending in the IDP but directionally, this is where we're going

- Combine "push notifications" with "security events"
- Update outgoing push notifications to be HTTP Push format
- Add "identity recycled" outgoing push event (LG-2502)
- Use iss-sub (not iss_sub) everywhere to be consistent with spec


mitchellhenke (approved):
👍🏼 👍🏼

PR #152, opened 2020-09-25
Update and rename attributes.md to  ID-dev-docs

nguyenduy19843:
Spot


zachmargolis:
Hi, thanks for the contribution. I don't think we want to rename this page at the moment, so I'm going to close out this PR. Please let us know if there's a specific issue you're looking to have fixed.

PR #158, opened 2020-12-23
[LG-3919] Add status page info to FAQ and footer

orenyk:
Resolves LG-3919

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -142,6 +142,9 @@
             <div class="grid-col-auto">
               <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
             </div>
+            <div class="grid-col-auto tablet:grid-col-fill text-right">
+              <a href="https://logingov.statuspage.io/">System Status</a>
+            </div>
           </div>
         </div>
       </div>
--- a/_pages/faq.md
+++ b/_pages/faq.md
@@ -13,8 +13,8 @@ How does login.gov work with my site?
 <div class="usa-accordion__content" markdown="1"  aria-expanded="true">
 We encourage you to create an account directly on [login.gov](https://secure.login.gov/) or an agency partner like
 [USAJobs](https://www.usajobs.gov/) to see login.gov in action.
-Generally a site will place a login button on their site.  When the user clicks this button they redirect to login.gov
-where they can sign in or create an account.  The login.gov site will be branded with the agency logo and can
+Generally a site will place a login button on their site. When the user clicks this button they redirect to login.gov
+where they can sign in or create an account. The login.gov site will be branded with the agency logo and can
 include help text for migrating existing users.
 After authenticating with login.gov they are redirected back to the agency with a unique UUID or email address that
 identifies the user.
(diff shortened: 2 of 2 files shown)


_layouts/base.html
```diff
@@ -142,6 +142,9 @@
             <div class="grid-col-auto">
               <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
             </div>
+            <div class="grid-col-fill text-right">
+              <a href="https://logingov.statuspage.io/" target="_blank">System Status</a>
```
aduth, line 146: Any strong attachment to opening in a new window?

https://www.w3.org/TR/WCAG20-TECHS/G200.html

```suggestion
              <a href="https://logingov.statuspage.io/">System Status</a>
```

_layouts/base.html
```diff
@@ -142,6 +142,9 @@
             <div class="grid-col-auto">
               <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
             </div>
+            <div class="grid-col-fill text-right">
```
aduth, line 145: Should we allow this to collapse to a new line in smaller viewports? It gets a bit tight. We can use [USWDS responsive class prefixes](https://designsystem.digital.gov/utilities/layout-grid/#responsive-classes) to customize it per-viewport range.

image

aduth (approved):

_layouts/base.html
orenyk: Great suggestion, fixed!

_layouts/base.html
orenyk: Hm, as an external link it felt like a new window made more sense, it's also what we're doing on the new brochure website: https://github.com/18F/identity-site/blob/demo/_includes/footer.html#L52. What do you think @aduth?

aduth: My preference would be to open it in the same tab. While opening external links in a new tab is a common practice, my understanding is that it's increasingly recommended to open in the same tab whenever possible, for mobile-friendliness and [improved accessibility](https://www.w3.org/TR/WCAG20-TECHS/G200.html) as mentioned above, and to avoid [recommended accommodations](https://www.w3.org/TR/WCAG20-TECHS/G201.html) associated with opening links in a new window.

There was some related discussion about this in https://github.com/18F/identity-idp/pull/4310, where in some cases we may have legitimate reason to open in new windows, but largely in our apps like the IdP, to avoid potential in-progress form data loss.

orenyk: Fair enough, I don't have a strong preference so I'll change it. Not sure if you want to make the same change on the main login.gov site as well.

aduth: > Fair enough, I don't have a strong preference so I'll change it. Not sure if you want to make the same change on the main login.gov site as well.

Yeah, it'd be good to be consistent. I can put together a pull request.

aduth: Following-up: I created LG-3976 to track this, since there are _many_ (100+) instances of links on the brochure site opening in new windows, and it should ideally be addressed holistically.

orenyk: Thanks for the update! That's way more instances than I would have expected, to be honest - glad you mentioned it!

PR #159, opened 2020-12-23
specify .txt

porta-antiporta:


aduth (commented):
Is the file extension strictly necessary? I'm not aware of where we'd validate this in the code, and I'd just tested in int with [this `.yml` file](https://github.com/18F/identity-idp/blob/master/spec/fixtures/ial2_test_credential_no_dob.yml) and it worked as expected.

One restriction I know of is that it must be assigned as the field for "Back Image" -- a hold-over from the pre-React implementation where it was its own step. I'm sure we could revise this to allow it to upload from either front or back.

aduth:
Closing as stale. To my knowledge and as mentioned in previous comment, there is no specific file extension required.

PR #163, opened 2021-02-09
Add public cert generation instructions

orenyk:


zachmargolis:
With this new PR, should we close out this old PR? https://github.com/18F/identity-dev-docs/pull/59

orenyk:
@zachmargolis good question - that one is way more comprehensive, do we think it's worth expanding to include the conf file and all that?

zachmargolis:
> that one is way more comprehensive, do we think it's worth expanding to include the conf file and all that?

If the abbreviated version is what we've been telling partners, and it works, seems good enough and less error-prone

achapm (approved):
👍  I like this version better because it is concise and not specific to oidc or saml. 
LGTM!

PR #165, opened 2021-02-17
Add SAML certs to SAML page

orenyk:
As discussed with a partner:

> ... I need an out-of-band source for the signing certificate.  An attacker who can change the SAML metadata can also change the embedded signing certificate.  If the signing certificate were published on a separate system (the developer docs is a pretty logical place for this, or sent with the production provisioning emails), an attacker would have a higher risk of exposure because they'd need to subvert two (theoretically separate) systems: the metadata distribution endpoint and the documentation library.

We'll have to update this when we do the 2021 cert rotation.

### DIFF
--- a/_pages/saml.md
+++ b/_pages/saml.md
@@ -66,15 +66,80 @@ Here are values needed to configure your service provider (SP) to work with logi
   ```
 
 - **x509 Public Certificate**
-  The public certificate is used to validate the authenticity of SAML requests received from login.gov, a minimum of 2048 bits. We publish this public certificate from in our [Metadata endpoint](#metadata).
+  The public certificate is used to validate the authenticity of SAML requests received from login.gov, a minimum of 2048 bits. We publish this public certificate from in our [metadata endpoint](#metadata) and below for verification.
 
 ### Metadata
 
-Consistent with the [SAML metadata specification](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf), login.gov's metadata is available at [https://idp.int.identitysandbox.gov/api/saml/metadata2020](https://idp.int.identitysandbox.gov/api/saml/metadata2020)
+Consistent with the [SAML metadata specification](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf), login.gov's metadata for our **sandbox** environment is available at [https://idp.int.identitysandbox.gov/api/saml/metadata2020](https://idp.int.identitysandbox.gov/api/saml/metadata2020).
 
-### Annual Certificate Rotation
+### Signing Certificates
+Below you can find the X509 certificates used by the login.gov IdP to sign SAML requests. **Do not enter these certificates in the Dashboard when configuring an application for testing** - you can follow the instructions in our [testing article]({% link _pages/testing.md %}#creating-a-public-certificate) to generate a client certificate.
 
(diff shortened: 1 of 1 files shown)


achapm (approved):
LGTM

_pages/saml.md
```diff
@@ -66,15 +66,68 @@ Here are values needed to configure your service provider (SP) to work with logi
...
-  The public certificate is used to validate the authenticity of SAML requests received from login.gov, a minimum of 2048 bits. We publish this public certificate from in our [Metadata endpoint](#metadata).
+  The public certificate is used to validate the authenticity of SAML requests received from login.gov, a minimum of 2048 bits. We publish this public certificate from in our [metadata endpoint](#metadata) and below for verification.
 
 ### Metadata
 
-Consistent with the [SAML metadata specification](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf), login.gov's metadata is available at [https://idp.int.identitysandbox.gov/api/saml/metadata2020](https://idp.int.identitysandbox.gov/api/saml/metadata2020)
+Consistent with the [SAML metadata specification](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf), login.gov's metadata for our **sandbox** environment is available at [https://idp.int.identitysandbox.gov/api/saml/metadata2020](https://idp.int.identitysandbox.gov/api/saml/metadata2020).
 
-### Annual Certificate Rotation
+### Signing Certificates
```
achapm, line 75: Let's add a note that this is the IDP certificate, not the SP certificate. To create the SP certificate "Go here" link to testing tab

achapm (approved):
LGTM pending note above

_pages/saml.md
```diff
@@ -66,15 +66,68 @@ Here are values needed to configure your service provider (SP) to work with logi
...
+The **2020** X509 certificate used to sign SAML requests for our **sandbox** environment (**idp.int.loginsandbox.gov**) can be found below (in addition to the metadata endpoint):
 
-The Login.gov SAML certificate is valid for just over one year. Every spring, Login.gov adds new SAML endpoints with the current year that use a new signing certificate. (So `/api/saml/metadata2020` becomes `/api/saml/metadata2021`.)
+```
```
zachmargolis, line 78: WDYT of `<summary>` `<detail>` tags so so that we can compress the cert bodies to make the page easier to skm?

orenyk: Love it, great suggestion. Thanks!

_pages/saml.md
orenyk: Done

zachmargolis: So it turns out this didn't render correctly.

And then I realized we already have accordions on this page ("View example authentication request" etc) can we just use those instead?

orenyk: @zachmargolis Ha saved by the bell, I was about to merge. I don't have strong preferences but I think I got the details/summary thing to work - this will be more concise I think. Any concerns with just merging as-is? ([preview here](https://federalist-14ccf384-8bd5-4e30-aa95-a5fe5894e3c1.app.cloud.gov/preview/18f/identity-dev-docs/oyk-saml-cert/saml/#signing-certificates))

zachmargolis: Preview looks good, merging as-is would be fine, but I think we should follow-up to unify the page and make it consistent with itself, so if that's accordion or details/summary I don't feel strongly

orenyk: Eh I'll just switch it - more likely to happen now than later 😄

PR #170, opened 2021-03-12
LG-4325 confirm the data formats for all IAL2 attributes

achapm:
WHY:
Confirm the data formats for all IAL2 attributes so that developers can identify any need for data transformation.

### DIFF
--- a/_pages/attributes.md
+++ b/_pages/attributes.md
@@ -6,6 +6,11 @@ lead: >
 
 Here are the possible attributes that can be requested at a given IAL. This table contains the available user attributes, the IAL they are associated with, and how they can be accessed in OpenID Connect and SAML.
 
+It is important to expect any number of characters in the `(string)` datatype unless directly followed by a number such as `(string36)`.
+
+The following fields are stored together in a text datatype with a maximum length of 65,535 bytes:<br>
+First name, Last name, Address, Phone, Date of birth, Social security number
+
 <table>
   <thead>
     <th>Attribute</th>
@@ -26,7 +31,7 @@ Here are the possible attributes that can be requested at a given IAL. This tabl
 ![checkmark][checkmark]
 </td>
 <td markdown="1">
-`sub` (string)
+`sub` (string36)
 </td>
 <td markdown="1">
 `uuid`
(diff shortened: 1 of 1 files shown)


_pages/attributes.md
```diff
@@ -142,14 +147,15 @@ Requires `profile` or `profile:birthdate` scopes.
     </tr>
     <tr>
 <td markdown="1">
-**Social security number**
+**Social security number**<br>
+Example: 111-11-1111
```
achapm, line 151:
```suggestion
Example:<br>`111-11-1111`
```

_pages/attributes.md
```diff
@@ -6,6 +6,11 @@ lead: >
 
 Here are the possible attributes that can be requested at a given IAL. This table contains the available user attributes, the IAL they are associated with, and how they can be accessed in OpenID Connect and SAML.
 
+It is important to expect any number of characters in the `(string)` datatype unless directly followed by a number such as `(string36)`.
+
+The following fields are stored together in a text datatype with a maximum length of 65,535 bytes:<br>
+First name, Last name, Address, Phone, Date of birth, Social security number
```
zachmargolis, line 12: Is this useful to people? That's also the length of the blob before it's encrypted, so I suspect the actual limit may be slightly less?

zachmargolis (approved):
LGTM, I think it's probably time we move beyond this tabular view of the attributes and to just a more consistent flow with a table of contents... the table is getting crowded

_pages/attributes.md
achapm, line 12: Yes, we will be re-working the dev docs site soon following the structure in the brochure and partners site.

What about saying this instead? I'm trying to be as transparent as possible.
```suggestion
It is important to expect any number of characters in the `(string)` datatype unless directly followed by a number such as `(string36)`. Strings are encrypted and stored in a text datatype with a maximum length of 65,535 bytes.
```

_pages/attributes.md
```diff
@@ -96,7 +101,7 @@ Requires `profile` or `profile:name` scopes.
 <td markdown="1">
 `address` (object)
 
-The [address claim](https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim), containing `street_address`, `locality` (city), `region` (state), and `postal_code` (zip code). Requires the `address` scope.
+The [address claim](https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim), containing `street_address`(string), `locality`(city, string), `region`(state, string), and `postal_code`(zip code, string5). Requires the `address` scope.
```
achapm, line 101: Make this easier to read:
```suggestion
The [address claim](https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim), containing: `street_address`(string) <br>`locality`(city, string) <br> `region`(state, string) <br> `postal_code`(zip code, string5) <br><br> Requires the `address` scope.
```

_pages/attributes.md
achapm: Thanks. See following revision.

orenyk (approved):
:shipit:

PR #178, opened 2021-04-14
Turn RISC events into a data file, export data/risc.json

zachmargolis:
A discussion with @juliaelman on RISC events in the dashboard turned into "wouldn't it be nice to show the pretty name of an event in the dashboard?"

So then I figured we could expose the data from the dev docs as a `.json` that the dashboard could load + cache

I also realized that now that we're adding a few more types of RISC events, so hopefully this could help us document there

Future ideas:
- Maybe instead of one long `/security` page, we break each event into its own page and deep link there?
- A field to link to the OpenID spec for particular RISC events

### DIFF
--- a/_data/risc.yml
+++ b/_data/risc.yml
@@ -1,0 +1,70 @@
+- friendly_name: Authorization Fraud Detected
+  direction: incoming
+  event_type: https://schemas.login.gov/secevent/risc/event-type/authorization-fraud-detected
+  description: >
+    RPs should submit this event when they believe a user's credentials have been compromised,
+    that somebody who is not the user was able to sign in to a user's account. Login.gov may force
+    the reset of user's password when we receive this event.
+- friendly_name: Identity Fraud Detected
+  direction: incoming
+  event_type: https://schemas.login.gov/secevent/risc/event-type/identity-fraud-detected
+  description: >
+    RPs should submit this event when they suspect an account has been used to commit identity theft or related fraudulent activity. Login.gov may reset the user's profile and verified attributes data when we receive this event.
+
+- friendly_name: Account Purged
+  direction: outgoing
+  event_type: https://schemas.openid.net/secevent/risc/event-type/account-purged
+  description: >
+    Login.gov pushes this event when a user deletes their account.
+  payload_schema:
+    - key: subject
+      description: An event should will a **subject** object, with the following keys
+      properties:
+        - key: subject_type
+          type: string
+          description: >
+            Will be **iss-sub**, this indicates the **sub** is the subject provided by the original issuer (login.gov)
+        - key: iss
+          type: string
+          description: >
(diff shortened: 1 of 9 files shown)


_pages/security-events.md
```diff
@@ -232,79 +227,36 @@ To configure your application to receive notifications from login.gov, supply lo
...
-  }
-}
+{{ event.example_payload | jsonify | pretty_jsonify }}
 ```
 
+{% endfor %}
+
 ### Request
 
 login.gov will make a POST request to your app's `push_notification_url`, see [Configuration](#configuration) for more details on setting that up. The JWT will be signed with login.gov's private key. See the OpenID Connect guide for information on how to get login.gov's public key from the [Certificates Endpoint](/oidc/#certificates).
```
juliaelman, line 262: s login.gov/Login.gov

juliaelman (changes requested):
Great work! One small content change.

_pages/security-events.md
zachmargolis: Great catch! Updated in d8b2384f4fd14fc6b0a4bafb2828e8a0a43d97a2

_data/risc.yml
```diff
@@ -0,0 +1,77 @@
+#
```
jmhooper, line 1: Since the format we use for incoming and outgoing events is slightly different, any thoughts on breaking those into separate files? e.g. `risc_outgoing.yml` and `risc_incoming.yml`. Corrects for the issue where someone adding an outgoing event copies the first one in the file and pastes it at the bottom and winds up missing a bunch of keys.

_includes/schema.html
```diff
@@ -0,0 +1,25 @@
...
+- schema, array of:
+    - key
+    - description
+    - type (optional)
+    - properties (optional)
+{% endcomment %}
+<ul>
+  {% for item in include.schema %}
+    <li>
+<div markdown="1">
```
jmhooper, line 12: Presuming this is not indented b/c of the unfortunate quirk where things are escaped if they are indented more than four spaces and included in a markdown file

jmhooper (approved):

_data/risc.yml
zachmargolis: Makes sense! Will give it a shot. Still combine them in the resulting `/data/risc.json`?

mitchellhenke: I like that idea.

_includes/schema.html
zachmargolis: yes :[

_data/risc.yml
zachmargolis: PTAL @ ba759b78fd0d7c3c28c473984a39c91ff26e357f

juliaelman (approved):
🚀

PR #180, opened 2021-04-15
Idea: Convert parts of the OIDC page to use a schema partial

zachmargolis:
**Why**: Code reuse for consistent formatting

This is just an idea, it might make things more complex than needed, but I wanted to give it a shot

A few other thoughts while I'm here...
- I think that this page is too long, and each endpoint should be broken out into its own page
- The authorization params section is definitely too complex to be a nested list, I think that we might want to break those into like list/headings/sections with deep links to each of those


_includes/schema.html
```diff
@@ -3,23 +3,24 @@
...
-  ({{ item.type }})
-{%- endif -%}
-<br />
-{{ item.description }}
-</div>
-      {% if item.properties %}
-        {% include schema.html schema=item.properties %}
-      {% endif %}
+    <li markdown="1">
+<!-- -->**{{ item.key }}**
```
zachmargolis, line 13: this is my fake indentation that lets maintain markdown formatting... I think I may just give up and not markdown it... changes incoming

zachmargolis: nevermind... this is the best way to get jekyll to not stick a giant `<p>` tag at the start of the description, which ends up adding way too much visual space

mitchellhenke (approved):
Agree that the page is long and could be broken up 👍🏼

stevegsa (approved):

zachmargolis:
Closing in favor of just doing separate pages in #183

PR #179, opened 2021-04-15
Document additional outgoing RISC event types:

zachmargolis:
- Identifier Changed
- Recovery Activated
- Recovery Information Changed

PR #181, opened 2021-04-19
Fix indentation to fix data type

zachmargolis:
Good catch, @aduth @ https://github.com/18F/identity-dev-docs/pull/179#discussion_r615825890


orenyk (approved):
:shipit:

PR #184, opened 2021-04-20
MFA Limit Account Locked and Re-proofing RISC events

stevegsa:


_data/risc_outgoing.yml
```diff
@@ -112,3 +129,20 @@
       }
     }
   }
+
+- friendly_name: Reproofing Completed
+  event_type: https://schemas.login.gov/secevent/risc/event-type/reproof-completed
+  spec_url:
+  description: >
+    Login.gov pushes this event when a user reproofs after having an active verified profile.
```
zachmargolis, line 137: I feel like "reproof" is a word we invented, and "proofing" is kind of jargony, what about something like this?

```suggestion
    Login.gov pushes this event when a user re-verifies their identity (goes through the identity proofing process). This event is only sent if they previously had an active verified profile, not for the first-time identity verification.
```

_data/risc_outgoing.yml
```diff
@@ -77,6 +77,23 @@
     }
   }
 
+- friendly_name: MFA Limit Reached Account Locked
```
zachmargolis, line 80: Let's expand the acronym? maybe flip the words around?

```suggestion
- friendly_name: Account Locked Due to MFA (Multi-Factor Authentication) Limit Reached
```

_data/risc_outgoing.yml
```diff
@@ -77,6 +77,23 @@
...
+  event_type: https://schemas.login.gov/secevent/risc/event-type/mfa-limit-account-locked
+  spec_url:
+  description: >
+    Login.gov pushes this event when a user hits the limit for failed MFA and gets locked out.
```
zachmargolis, line 84: expand the acronym

```suggestion
    Login.gov pushes this event when a user fails MFA (Multi-Factor Authentication) with their second factor enough times to hit the rate limit, and is subsequently locked out of their account for a period of time.
```

zachmargolis (approved):
Some wording suggestions, otherwise LGTM

PR #205, opened 2021-11-16
Add the all emails scope attribute for SAML

jmhooper:
**Why**: So that partners will know how this works


zachmargolis (approved):
LGTM

orenyk (approved):
:shipit:

PR #208, opened 2021-12-10
Add notice of change in allowed SSNs in sandbox testing

mitchellhenke:
Will follow up with updates to the language closer to the implementation of the change.


orenyk (approved):
:shipit:

PR #222, opened 2022-01-28
Hot swap certs

jmhooper:
**Why**: The previous certs were generated with invalid before and after dates


zachmargolis (approved):
LGTM

orenyk (approved):
Verified staging and prod, one comment re: noting that this will be available 2/3

PR #235, opened 2022-02-23
Fix language around client authentication vs PKCE

aaronpk:
PKCE is not a form of client authentication and shouldn't be referred to as such. The only form of client authentication supported by login.gov is `private_key_jwt`.

Also fixed the PKCE code_challenge sample code.


_pages/oidc.md
```diff
@@ -36,17 +36,18 @@ sidenav:
 
 ## Getting started
 
-### Choosing an authentication method
+### Client authentication
 
-Login.gov supports two ways of authenticating clients: **private_key_jwt** and **PKCE**.
+Login.gov supports only **private_key_jwt** for client authentication.
```
zachmargolis, line 41: if PKCE is not considered client authentication, what is the name of the umbrella category where we can offer both private_key_jwt and PKCE as options to integrate with us?

_pages/oidc.md
```diff
@@ -36,17 +36,18 @@ sidenav:
...
+**client_secret** is not supported by Login.gov because of security reasons. It requires managing a shared secret in two places: the client and the server, where **private_key_jwt** flow involves only sharing public keys with the server.
 
-Other implementations of OpenID Connect use the "implicit flow" or the **client_secret** param, but Login.gov does not support those methods. The implicit flow is [not recommended by the OAuth group][oauth-implicit-deprecated] for security reasons. Similarly **client_secret** is not supported by Login.gov because of security reasons. It requires managing a shared secret in two places: the client and the server, where **private_key_jwt** flow involves only sharing public keys with the server and **PKCE** only has a one-time secret.
+Clients that are unable to use **private_key_jwt** (e.g. native mobile apps) must use PKCE, short for [Proof Key for Code Exchange by OAuth Public Clients](https://tools.ietf.org/html/rfc7636) and pronounced "pixy." In this method, the client sends a public identifier as well as a hashed random value generated by the client.
```
zachmargolis, line 48: I think this worked better as an item in the bulleted list, since it's our alternative option to private key jwt. Also for this line, I think the emphasis should be on PKCE rather than jwt:

```suggestion
Clients that are unable to use private_key_jwt (e.g. native mobile apps) must use **PKCE**, short for [Proof Key for Code Exchange by OAuth Public Clients](https://tools.ietf.org/html/rfc7636) and pronounced "pixy." In this method, the client sends a public identifier as well as a hashed random value generated by the client.
```

_pages/oidc.md
aaronpk: I tried to keep this PR simple and to describe the current behavior accurately. PKCE is also recommended even if you are using any form of client authentication including private_key_jwt, but that's a different discussion to be had 😄  This PR doesn't attempt to change any behavior of the current system, just avoids referring to PKCE as a form of client authentication. If you wanted to have two options side by side, then I would say the forms of client authentication are **private_key_jwt** and **none**, where PKCE is required if you are not including client authentication.

_pages/oidc.md
aaronpk: The change was based on the idea that PKCE is explicitly not a form of client authentication, and including it as an alternative to private_key_jwt makes it look like it is.

_pages/oidc.md
zachmargolis: Ah, I see. We'll probably go with that private_key_jwt/none approach then, and keep the bulleted list

aaronpk: Alright, I updated the PR to bring back the bullet list.

_pages/oidc.md
```diff
@@ -36,17 +36,16 @@ sidenav:
...
 
-### Choosing an authentication method
+### Client authentication
 
-Login.gov supports two ways of authenticating clients: **private_key_jwt** and **PKCE**.
+Login.gov supports **private_key_jwt** client authentication or no authentication.
 
 - **private_key_jwt** (preferred for **web apps**)
   The client sends a [JSON Web Token][jwt], or JWT, signed with a private key (minimum length of 2048 bits) when requesting access tokens. The corresponding public key is registered with the IdP ahead of time, similar to SAML.
+- **none** (for **native mobile apps**)
```
zachmargolis, line 45: I see how this is technically correct, but I think in terms of implementation, it may be misleading and confusing based on the rest of the page, and our expected workflows

Since the rest of the page has toggles for private_key_jwt/PKCE (see example below) I think this might be preferable?

```suggestion
- *none* with **PKCE** (for **native mobile apps**)
```

| example |
| --- |
|  Screen Shot 2022-02-23 at 11 58 11 AM |

aaronpk: alright, I can live with that. I'm going to get in touch on HackerOne about something related though. Do you see those reports?

zachmargolis: Yes, we do see the HackerOne reports

_pages/oidc.md
```diff
@@ -36,17 +36,16 @@ sidenav:
 
 ## Getting started
 
-### Choosing an authentication method
+### Client authentication
```
orenyk, line 39: **suggestion:** The nuance of `client authentication` may ultimately confuse our audience more than clarify things - can we just refer to these as `OIDC flows`?
```suggestion
### Choosing an OIDC flow
```

_pages/oidc.md
aaronpk: I just tried and it says the program is disabled. I can just email instead

_pages/oidc.md
```diff
@@ -36,17 +36,16 @@ sidenav:
...
 
-Login.gov supports two ways of authenticating clients: **private_key_jwt** and **PKCE**.
+Login.gov supports **private_key_jwt** client authentication or no authentication.
```
orenyk, line 41:
```suggestion
Login.gov supports two OIDC flows: **private_key_jwt** and **PKCE**.
```

_pages/oidc.md
```diff
@@ -36,17 +36,16 @@ sidenav:
...
 
 - **private_key_jwt** (preferred for **web apps**)
   The client sends a [JSON Web Token][jwt], or JWT, signed with a private key (minimum length of 2048 bits) when requesting access tokens. The corresponding public key is registered with the IdP ahead of time, similar to SAML.
+- *none* with **PKCE** (for **native mobile apps**)
```
orenyk, line 45: **suggestion:** the `none` phrasing seems like it could be similarly confusing:
```suggestion
- **PKCE** (preferred for **native mobile apps**)
```

_pages/oidc.md
zachmargolis: Ah thanks for letting us know, we'll fix the link on our security page, but in the meantime here is the right HackerOne project: https://hackerone.com/gsa_bbp?type=team

orenyk (changes requested):
Thanks for your input! I had some thoughts around keeping this more "plain language" - what do you think?

aaronpk:
I see what you're getting at, but I'm concerned that referring to `private_key_jwt` as a flow might also be confusing.

aaronpk:
Alright, let me explain the motivations here in more detail. At a high level, what I was trying to do with this was to avoid putting PKCE in a list alongside client authentication, since it is not client authentication.

PKCE is one way to prevent authorization code injection attacks, the other is `nonce`. PKCE is enforced by the authorization server, `nonce` prevents authorization code injection attacks if checked by the client (although the attack would have already happened at that point, and ID token issued, but the client can then abort when it detects the attack has happened).

Authorization code injection attacks are still possible with client authentication, including `private_key_jwt` authentication. That is my concern with listing it as an alternative to `private_key_jwt`.

I've decided it's okay to talk about this here instead of privately via HackerOne because login.gov does require clients send the `nonce` parameter, so it's doing the right thing and it's up to the client to prevent authorization code injection attacks, it's not a vulnerability with login.gov. That said, I would still rather see login.gov require PKCE for all clients, since it is then possible for the AS to prevent the authorization code injection in the first place, rather than wait for the client to reject it and hope the client implements that properly.

So there are really two separate issues:

* Avoid making it look like PKCE is an alternative to client authentication
* Require (or at least make possible) PKCE even if a client uses `private_key_jwt` authentication

ReyBan78 (approved):

aaronpk:
Just wanted to follow up on this again. Any chance we can at least get the language on the docs fixed?

Longer term I still strongly recommend that you require (or at least make possible) using PKCE even when using `private_key_jwt` authentication, since it prevents authorization code injection and is enforced by the authorization server. Relying on the client developer to use and validate `nonce` to prevent authorization code injection is more risky.

zachmargolis:
Hi, thanks again for submitting this.

We agree with a preference for JWT over PKCE, but don't think this particular document is the right place to deep dive on those nuances that depend on the structure of partner applications and their threat models. We have additional communications with our partners, and this page outlines our capabilities as a reference document.

We'll close this for the time being, and as we update our documentation, we'll consider adding adding additional guidance.

aaronpk:
Sorry if this got kind of longwinded and covering a larger issue, but this is still a problem with the docs. 

The core of the issue is that **PKCE is not a form of client authentication**. The sentence "Login.gov supports two ways of authenticating clients: private_key_jwt and PKCE." is incorrect, since PKCE is not client authentication. 

The policy around what kinds of apps require PKCE or use private_key_jwt authentication is a separate (and interesting) issue, but not relevant to the documentation issue. PKCE can and should be used along with private_key_jwt authentication, but that's an actual product question and I understand that you don't want to change that behavior right now.

So very specifically, the change I would like to see is to avoid referring to PKCE as a form of client authentication. The smallest changes to accomplish this would be:

* Change the header from "Choosing an authentication method" to "Choosing an OIDC flow"
* Change "Login.gov supports two ways of authenticating clients: private_key_jwt and PKCE." to "Login.gov supports two OIDC flows: private_key_jwt and PKCE."

While it's somewhat misleading to refer to private_key_jwt and PKCE as separate OIDC flows, it's less misleading than calling PKCE a form of client authentication.

Even better would be the more clear separation of client authentication and PKCE that I had originally suggested in the PR, but I understand if that's too much of a change.

PR #238, opened 2022-03-08
[LG-5881] Add Zendesk banner

orenyk:
Also tweaked some headers on the support page for consistency with the rest of the site

### DIFF
--- a/_pages/index.md
+++ b/_pages/index.md
@@ -15,6 +15,8 @@ layout: home
 
 <section class="usa-section grid-container usa-prose" markdown="1">
 
+{% include zendesk-alert.html %}
+
 # Get started...
 
 - [Understand our flow]({{ site.baseurl }}/overview/).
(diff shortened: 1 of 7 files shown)


_pages/index.md
```diff
@@ -15,6 +15,20 @@ layout: home
 
 <section class="usa-section grid-container usa-prose" markdown="1">
 
+{%- capture alert_content -%}
```
zachmargolis, line 18: we could probably make one partial for this and use it on all these pages, and then include that everywhere? Like one partial specifically for this migration banner that itself includes the alert partial?

zachmargolis (approved):
LGTM

_pages/index.md
orenyk: yeah that would probably make even more sense. Let me take a stab at that

zachmargolis (approved):

PR #244, opened 2022-05-02
fix PKCE example

aaronpk:
PKCE requires trimming the trailing "=" from the Base64 encoding 

https://datatracker.ietf.org/doc/html/rfc7636#appendix-A


zachmargolis (approved):
LGTM, thank you for helping keep our examples accurate!

PR #247, opened 2022-05-11
Add codeowners

NatashaJIbrahim:
This will ensure that @18F/identity-ursula is always tagged for review and that pr's are looked at in a timely manner.


orenyk (approved):
:shipit:

PR #250, opened 2022-05-20
Remove zendesk-alert

NatashaJIbrahim:
Removes the zendesk alert since it's a bit dated now and traffic to Hubspot has slowed down.


zachmargolis (approved):
LGTM

PR #255, opened 2022-06-15
clarify that a real phone number is needed for testing

afeld:
I tried entering a fake phone number and got locked out; didn't understand that it was actually going to text/call! Still not entirely clear to me what the difference between the sandbox and production are in this regard; may want to strike the previous sentence, or further clarify. Thanks!

### DIFF
--- a/_pages/testing.md
+++ b/_pages/testing.md
@@ -151,7 +151,7 @@ To simulate a failure, enter a social security number that does not start with â
 
 ### Phone number verification
 
-Login.gov collects a phone number during the proofing process. In a live production environment, Login.gov checks that this phone number is associated with the applicant. You can use any phone number for testing purposes in the sandbox environment other than the following:
+Login.gov collects a phone number during the proofing process. In a live production environment, Login.gov checks that this phone number is associated with the applicant. You can use any real phone number for testing purposes in the sandbox environment other than the following:
 
 * `703-555-5555` - simulates a phone number that couldn't be verified as belonging to the user
 * `703-555-5888` - simulates a timeout during verification


NatashaJIbrahim (commented):
@afeld Thank you for this! I agree this should be made more clear. My understanding is that in Production we check that the phone is associated with the address of the user during Identity Proofing. However, in Sandbox this check doesn't happen since it depends on a third party service. This check also doesn't happen for basic authentication. 

I'm sorry you were locked out of your account. If it's any consolation, the Sandbox and Production accounts are discrete. One trick we use here when testing (that you might already be familiar with) is to append the date to the end of our email address for example, "my-email@example.com" becomes "my-email+06152022@example.com". They'll be treated as the same address by most email services but as different addresses by Login. This can allow you to have multiple testing accounts in the Login Sandbox environment. 

I'll look over this one more time in the morning.

julialeague:
You can also get away with using a fake phone number, as long as you're ok with logging in with the generated backup codes for that user. Understandable if you don't! That could be clarified somewhere.

_pages/testing.md
```diff
@@ -151,7 +151,7 @@ To simulate a failure, enter a social security number that does not start with 
 
 ### Phone number verification
```
Richalonso45, line 152: 3523176827

Richalonso45, line 152: 593635325

_pages/testing.md
```diff
@@ -151,7 +151,7 @@ To simulate a failure, enter a social security number that does not start with 
...
 
-Login.gov collects a phone number during the proofing process. In a live production environment, Login.gov checks that this phone number is associated with the applicant. You can use any phone number for testing purposes in the sandbox environment other than the following:
+Login.gov collects a phone number during the proofing process. In a live production environment, Login.gov checks that this phone number is associated with the applicant. You can use any real phone number for testing purposes in the sandbox environment other than the following:
```
Richalonso45, line 154: 3523176827

_pages/testing.md
Richalonso45: 593635325

Richalonso45 (approved):

_pages/testing.md
Richalonso45: 3523176827

orenyk:
This will be resolved in a pending update to this page.

PR #258, opened 2022-07-19
URL encode SAML request payload

mmagsa:
The base64 encoding character set includes characters that are not URL-safe, so the base64-encoded payload should also be url-encoded.

### DIFF
--- a/_pages/saml.md
+++ b/_pages/saml.md
@@ -235,10 +235,10 @@ To authenticate a user with Login.gov, direct them to our authentication URL wit
 https://idp.int.identitysandbox.gov/api/saml/auth{{ page.saml_year }}?SAMLRequest=${SAML_REQUEST}
 ```
 
-The `SAMLRequest` parameter is a base64-encoded, deflate-compressed XML payload of a `<samlp:AuthnRequest>`:
+The `SAMLRequest` parameter is a url-encoded, base64-encoded, deflate-compressed XML payload of a `<samlp:AuthnRequest>`:
 
 ```
-SAML_REQUEST = base64(deflate(payload))
+SAML_REQUEST = urlEncode(base64(deflate(payload)))
 ```
 
 <div class="usa-accordion--bordered">


orenyk (approved):
:shipit:

_pages/saml.md
```diff
@@ -235,10 +235,10 @@ To authenticate a user with Login.gov, direct them to our authentication URL wit
 https://idp.int.identitysandbox.gov/api/saml/auth{{ page.saml_year }}?SAMLRequest=${SAML_REQUEST}
 ```
 
-The `SAMLRequest` parameter is a base64-encoded, deflate-compressed XML payload of a `<samlp:AuthnRequest>`:
+The `SAMLRequest` parameter is a url-encoded, base64-encoded, deflate-compressed XML payload of a `<samlp:AuthnRequest>`:
 
 ```
-SAML_REQUEST = base64(deflate(payload))
+SAML_REQUEST = urlEncode(base64(deflate(payload)))
```
zachmargolis, line 241: URL-safe base64 is more like a separate alphabet of base64 than just URL-escaping a base64 payload

some libraries do base64 urlsafe in one method, like Ruby: https://ruby-doc.org/stdlib-2.4.0/libdoc/base64/rdoc/Base64.html#method-i-urlsafe_encode64

so maybe we encourage folks to look up a method like that?

```suggestion
SAML_REQUEST = urlSafeBase64(deflate(payload)))
```

zachmargolis: They have different alphabets: https://base64.guru/standards/base64url

Here is an example that highlights different results for the same raw data:
```
require 'securerandom'
require 'cgi'
require 'base64'

bytes = SecureRandom.bytes(12)
=> "a\xB9\x1Fj\xAB\xBE\x96!N\nR+"
# plain base64
Base64.encode64(bytes).chomp
=> "Ybkfaqu+liFOClIr"
CGI.escape(Base64.encode64(bytes).chomp)
=> "Ybkfaqu%2BliFOClIr"
# url-safe base64
Base64.urlsafe_encode64(bytes)
=> "Ybkfaqu-liFOClIr"
```

orenyk: @zachmargolis doesn't our IdP expect regular base64 though? https://github.com/18F/saml_idp/blob/main/lib/saml_idp/request.rb#L12

zachmargolis: oh I guess so! 🤷

PR #260, opened 2022-07-27
Add int maintenance alert to all pages

orenyk:
This can be turned on by specifying the timeframe in _config.yml and
turned off by setting the `int_maintenance_window` configuration value
to `false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 [View it live ❯](https://developers.login.gov/)
 
+## Temporary Alert
+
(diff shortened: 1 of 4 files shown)


NatashaJIbrahim (approved):
<3

NatashaJIbrahim (commented):
Just thought of one slight improvement ... just to add this to the README. I can do that if you don't get to it.

orenyk:
@NatashaJIbrahim I refactored to make it more general (you can set any arbitrary site-wide alert) and added info to the README. Will merge once the build passes!

PR #265, opened 2022-08-08
Updating a field comment that was copy-pasted across two events

zachmargolis:
| | identifier-changed | identifier-recycled |
|---|----|---|
| email added | ✅ |  |
| email deleted | ✅* | ✅ |

So [looking at the code](https://github.com/18F/identity-idp/blob/1c6d88911e15edcf5ae92443c325324a00611018/app/forms/delete_user_email_form.rb#L28-L33) we send "identifier-changed" when an email is added and deleted? but i think it would make sense if we only did one? This standard is tough lol

### DIFF
--- a/_data/risc_outgoing.yml
+++ b/_data/risc_outgoing.yml
@@ -39,7 +39,7 @@
   spec_url: "https://openid.net/specs/openid-risc-event-types-1_0-ID1.html#rfc.section.2.6"
   description: >
     Login.gov pushes this event when a user removes an email address from their account, freeing up the email address as an identifier.
-  payload_schema: &email_schema
+  payload_schema:
     - key: subject
       description: An event should will a **subject** object, with the following keys
       properties:
@@ -67,7 +67,17 @@
   description: >
     Login.gov pushes this event when a user changes the email address on their account.
   payload_schema:
-    *email_schema
+    - key: subject
+      description: An event should will a **subject** object, with the following keys
+      properties:
+        - key: subject_type
+          type: string
+          description: >
+            Will be **email**
+        - key: email
+          type: string
+          description: >
+            The new email address that was added, or the the email address that was removed.
   example_payload: {
     "https://schemas.openid.net/secevent/risc/event-type/identifier-changed": {
       "subject": {


_data/risc_outgoing.yml
```diff
@@ -67,7 +67,17 @@
...
+      description: An event should will a **subject** object, with the following keys
+      properties:
+        - key: subject_type
+          type: string
+          description: >
+            Will be **email**
+        - key: email
+          type: string
+          description: >
+            The new email address that was added, or the the email address that was removed.
```
orenyk, line 80: **comment:** I don't think this is right, based on the [spec](https://openid.net/specs/openid-risc-event-types-1_0-ID1.html#rfc.section.2.5) the `email` key should be the old value and there should be another key at the same level as `subject` called `new-value` that contains the new email that was added.

That said, since we allow users to include multiple emails I'm actually not sure how to handle that case (since no email is removed). Maybe `email` can be null? @jmhooper @mitchellhenke any thoughts?

Also, it looks like we're using `Identifier Recycled` incorrectly - the spec says (emphasis mine):

> Identifier Recycled signals that the identifier specified in the subject was recycled **and now it belongs to a new user**.

This doesn't seem relevant to us until someone adds an email that previously belonged to someone else (or I guess deletes and recreates their account). Should we even be sending it at all?

Beyond that, there's this whole section in the `Identifier Changed` specification that makes me think we should only be using `Recovery Information Changed` for email updates (and that event does not include the email itself).

> This event SHOULD be issued only by the provider that is authoritative over the identifier. For example, if the person that owns john.doe@example.com goes through a name change and wants the new john.row@example.com email then only the email provider example.com SHOULD issue an Identifier Changed event as shown in the example below.
>
> If an identifier used as a username or recovery option is changed, at a provider that is not authoritative over that identifier, then [Recovery Information Changed](https://openid.net/specs/openid-risc-event-types-1_0-ID1.html#recovery-information-changed) [recovery-information-changed] SHOULD be used instead.

orenyk (commented):
This standard is tough! Somehow I think this opened more questions than it answered 😭

_data/risc_outgoing.yml
orenyk: cc @NatashaJIbrahim @julialeague @mmagsa @nprimak for awareness

hnji: Lun frla

hnji: Hii

_data/risc_outgoing.yml
```diff
@@ -39,7 +39,7 @@
   spec_url: "https://openid.net/specs/openid-risc-event-types-1_0-ID1.html#rfc.section.2.6"
   description: >
     Login.gov pushes this event when a user removes an email address from their account, freeing up the email address as an identifier.
-  payload_schema: &email_schema
```
hnji, line 42: Hnjikidaa

_data/risc_outgoing.yml
```diff
@@ -67,7 +67,17 @@
   description: >
     Login.gov pushes this event when a user changes the email address on their account.
   payload_schema:
-    *email_schema
```
hnji, line 70: Saabmultani502@gmail.com

_data/risc_outgoing.yml
hnji: Kamalohi20@gmail.com

hnji, line 42: Kamalohi20@gmail.com

hnji (approved):
Hnji

zachmargolis:
Seems like we never got to consensus on this, and it's stalled, so I'm going to close it.

PR #269, opened 2022-09-16
Make YAML files linkable for download

aduth:
**Why**: For easier reference in an external context (e.g. Slack, Google Doc).

### DIFF
--- a/_includes/yaml_download.md
+++ b/_includes/yaml_download.md
@@ -4,15 +4,16 @@ Include:
 - filename
 {% endcomment %}
 
+{% assign path = include.filename | prepend: 'yaml/' %}
+
+{%- capture data -%}
+{% include {{ path }} %}
+{%- endcapture %}
+
 ```yaml
-{{ include.data }}
+{{ data }}
 ```
 
-{% assign escaped_data = include.data | uri_escape | replace: '#', '%23' %}
-{% assign href = 'data:application/x-yaml,' | append: escaped_data %}
-
-<a href="{{ href }}"
-   download="{{ include.filename }}"
-   class="usa-button usa-button--outline">
+<a href="/assets/{{ path }}" class="usa-button usa-button--outline">
   Download .yml
 </a>
--- a/_plugins/yaml_content_disposition.rb
+++ b/_plugins/yaml_content_disposition.rb
@@ -1,0 +1,16 @@
+require 'webrick'
+
+module Jekyll
+  module Commands
+    class Serve
+      class Servlet < WEBrick::HTTPServlet::FileHandler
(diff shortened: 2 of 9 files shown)


aduth:
Example from the Federalist preview: https://federalist-14ccf384-8bd5-4e30-aa95-a5fe5894e3c1.app.cloud.gov/preview/18f/identity-dev-docs/aduth-yaml-content-disposition/assets/yaml/proofing.yml

_plugins/yaml_content_disposition.rb
```diff
@@ -0,0 +1,16 @@
+require 'webrick'
+
+module Jekyll
+  module Commands
+    class Serve
+      class Servlet < WEBrick::HTTPServlet::FileHandler
+        prepend(Module.new do
+          def do_GET(req, res)
+            res.header.merge!('Content-Disposition' => 'attachment') if req.path.end_with?('.yml')
```
aduth, line 9: Jekyll [supports custom headers](https://jekyllrb.com/docs/configuration/webrick/), but not for specific file types. This plugin follows the pattern from [the code which introduced that option](https://github.com/jekyll/jekyll/pull/4224/files#diff-7a28653629164aecdb4936731e0b70992db66523b2f8849a6295715326689d7f) to target YAML files specifically.

_includes/yaml_download.md
```diff
@@ -4,15 +4,16 @@ Include:
 - filename
```
zachmargolis, line 4: we should remove `- data` from the list of includes, since now that's an `assign`

_plugins/yaml_content_disposition.rb
```diff
@@ -0,0 +1,16 @@
...
+            super
+          end
+        end)
```
zachmargolis, line 12: not a super big fan of `end)` but this seems like the simplest overall, and we aren't going to touch this file that often

other ideas:

too sneaky:
```ruby
Module.new do
  # ...
end.tap { |mod| prepend mod }
```
more verbase
```ruby
attachment_mod = Module.new
  # ...
end

prepend attachment_mod
```

zachmargolis (approved):
LGTM

_includes/yaml_download.md
aduth: Good catch! I'll remove.

_plugins/yaml_content_disposition.rb
aduth: Yeah, and for context this is the form we've taken with other plugins which are monkeypatching Jekyll internals ([example](https://github.com/18F/identity-handbook/blob/main/_plugins/content_typography.rb)).

PR #275, opened 2022-10-11
Add state_id_jurisdiction to sample yaml (LG-7783)

mitchellhenke:


benjaminchait (approved):

mdiarra3 (approved):

aduth:
Could we change the IdP mock proofer to allow the value to be omitted?

mitchellhenke:
> Could we change the IdP mock proofer to allow the value to be omitted?

My gut reaction is I don't think so because the mock proofer and real proofers pass the parsed results to the same form for validation.  Adding a default was also mentioned [here](https://gsa-tts.slack.com/archives/C0NGESUN5/p1665514739516049?thread_ts=1665513387.098649&cid=C0NGESUN5), but my not particularly strong feeling is that it could be less intuitive in different ways by returning success for an otherwise invalid response.

PR #276, opened 2022-10-14
Add x509-related scope values

mmagsa:
https://developers.login.gov/attributes/ refers to x509-related scope values. These should be in the list of possible values.


julialeague (approved):
Nice catch!

PR #285, opened 2022-11-09
Add SHA256 requirement for auth request signatures

mmagsa:
Almost 30% of our SAML integrations currently send invalid (likely using SHA1 instead of SHA256) signatures. Specifically calling out this requirement in our documentation will hopefully prevent this in the future.

### DIFF
--- a/_pages/saml.md
+++ b/_pages/saml.md
@@ -291,6 +291,8 @@ An example authentication request, with indentation added for readability.
 </div>
 </div>
 
+**Note:** We strongly encourage you to cryptographically sign authentication requests, unless there are technical limitations on your side that prevent this. In addition to providing increased security, this allows for seamless rotation of your application's public certificate in the future. All signatures must use RSA SHA-256.
+
 ### Specifying attributes and assurance levels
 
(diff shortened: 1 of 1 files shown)


_pages/saml.md
```diff
@@ -291,6 +291,8 @@ An example authentication request, with indentation added for readability.
 </div>
 </div>
 
+**Note:** We strongly encourage you to cryptographically sign authentication requests, unless there are technical limitations on your side that prevent this. In addition to providing increased security, this allows for seamless rotation of your application's public certificate in the future. All signatures must use RSA SHA-256.
```
zachmargolis, line 294: is there a specific parameter or XML tag that this shows up in? as a hint for partners to know what to add to their requests?

zachmargolis: I was trying to put myself in a developer's shoes of "how can I check that I am sending this correctly" but yeah seems like a good backlog item if it's not quick to add here

zachmargolis (approved):
LGTM

julialeague (approved):
:shipit:

PR #289, opened 2022-11-29
Remove anchor visual bug on FAQ page (LG-7973)

zachmargolis:
- Removes bug by switching to `<div>` which don't get auto-anchored

Fixes 18F/identity-site#978

| before | after |
| --- | --- |
| Screen Shot 2022-11-29 at 9 56 45 AM | Screen Shot 2022-11-29 at 9 56 49 AM |

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -157,8 +157,7 @@
         '#main-content h1:not(.usa-accordion__heading)',
         '#main-content h2:not(.usa-accordion__heading)',
         '#main-content h3:not(.usa-accordion__heading)',
-        '#main-content h4:not(.usa-accordion__heading)',
-        '#main-content [data-anchor]'
+        '#main-content h4:not(.usa-accordion__heading)'
       ].join(', '));
     </script>
 
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -153,7 +153,13 @@
     <script src="{{ site.baseurl }}/assets/js/main.js"></script>
     <script src="{{ site.baseurl }}/assets/js/anchor.js"></script>
     <script>
-      anchors.add('#main-content h1:not(.usa-accordion__heading), #main-content h2, #main-content h3, #main-content h4, #main-content [data-anchor]');
+      anchors.add([
+        '#main-content h1:not(.usa-accordion__heading)',
+        '#main-content h2:not(.usa-accordion__heading)',
+        '#main-content h3:not(.usa-accordion__heading)',
+        '#main-content h4:not(.usa-accordion__heading)',
+        '#main-content [data-anchor]'
+      ].join(', '));
     </script>
 
     {% include dap.html %}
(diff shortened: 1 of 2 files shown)


aduth:
Poking around a bit, I see we're already trying to exclude some accordion headings:

https://github.com/18F/identity-dev-docs/blob/34692dcd0396df6c1a65d9740a16e1574e48b59e/_layouts/base.html#L156

Could we consider just enhancing this selector? Alternatively, if we want to switch to `div`, maybe we do so consistently and remove that handling in the `Anchors#add`? The more I think on it, the more heading might be semantically appropriate.

Maybe we could dynamically build the selector if we don't want to repeat all the same info, but it is a bit cryptic:

```js
anchors.add(Array(6).fill().map((_, i) => `#main-content h${i + 1}:not(.usa-accordion__heading)`).join());
```

zachmargolis:
> Could we consider just enhancing this selector?

What about adding `, :not([data-no-anchor])` to be the opposite of `[data-anchor]`?

aduth:
> What about adding `, :not([data-no-anchor])` to be the opposite of `[data-anchor]`?

Could you give a full example? Do you mean adding `data-no-anchor` attributes to the accordion headings and excluding that in the selector?

zachmargolis:
I just 

> > What about adding `, :not([data-no-anchor])` to be the opposite of `[data-anchor]`?
> 
> Could you give a full example? Do you mean adding `data-no-anchor` attributes to the accordion headings and excluding that in the selector?

I just realized that adding a single `:not()` will not negate each line 🤦 ... I am going to add the `:not(.usa-accordion__heading)` to each one, but also add some newlines, stand by

zachmargolis:
@aduth PTAL @ 6e478eaab0439ef0925248aa9af97d740eb24598

_layouts/base.html
```diff
@@ -153,7 +153,13 @@ <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
     <script src="{{ site.baseurl }}/assets/js/main.js"></script>
     <script src="{{ site.baseurl }}/assets/js/anchor.js"></script>
     <script>
-      anchors.add('#main-content h1:not(.usa-accordion__heading), #main-content h2, #main-content h3, #main-content h4, #main-content [data-anchor]');
+      anchors.add([
+        '#main-content h1:not(.usa-accordion__heading)',
+        '#main-content h2:not(.usa-accordion__heading)',
+        '#main-content h3:not(.usa-accordion__heading)',
+        '#main-content h4:not(.usa-accordion__heading)',
+        '#main-content [data-anchor]'
```
aduth, line 161: Existed previously, but I couldn't find anywhere this is used? Could we remove?

aduth (approved):
LGTM 👍

_layouts/base.html
zachmargolis: I guess so, we can always bring it back when we need it. removed in 909b25d65f13c7638eae50675047dec5faa81778

PR #297, opened 2023-01-24
Remove banner, corresponding update content was removed in #271

zachmargolis:
small follow up to #271. this banner is still up but the corresponding update info has been removed

Screen Shot 2023-01-24 at 1 12 59 PM


orenyk (approved):
yikes good catch, thanks!

PR #303, opened 2023-02-22
Fix bottom banner text color

nprimak:
After adding the survey to production, Annie noticed that the bottom banner had the wrong text color. This is a quick CSS fix to change the color. (see screenshots below) 

Screen Shot 2023-02-22 at 10 18 39 AM
Screen Shot 2023-02-22 at 10 18 27 AM

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -9,3 +9,12 @@
 #nav-links {
     width: fit-content !important;
 }
+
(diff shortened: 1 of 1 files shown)


orenyk (approved):
**question:** is this not configurable within Touchpoints? Assuming not, looks good to me!

PR #304, opened 2023-03-01
Upgrade ruby version

zachmargolis:
- Matches identity-site repo version


zachmargolis:
> Could we update to 3.2 and do the same in identity-site?

Sure! Will give it a shot

PR #306, opened 2023-03-10
Add markdown parsing for lead section

mmagsa:
Links and other markdown are currently not parsed in this section. Thanks @aduth for suggesting this fix!

### DIFF
--- a/_layouts/documentation.html
+++ b/_layouts/documentation.html
@@ -43,7 +43,7 @@ layout: base
         <div class="usa-display">{{ page.title }}</div>
 
         {% if page.lead %}
-          <p class="usa-intro">{{ page.lead | markdownify }}</p>
+          <div class="usa-intro">{{ page.lead | markdownify }}</div>
         {% endif %}
 
         {{ content }}


_layouts/documentation.html
```diff
@@ -26,7 +26,7 @@
 
           <div class="usa-display">{{ page.title }}</div>
 
-          {% if page.lead %}
+          {% if page.lead | markdownify %}
```
aduth, line 29: There's another reference below on line 46 we should probably update as well, at least for consistency?

_layouts/documentation.html
```diff
@@ -27,7 +27,7 @@
           <div class="usa-display">{{ page.title }}</div>
 
           {% if page.lead %}
-            <p class="usa-intro">{{ page.lead }}</p>
+            <p class="usa-intro">{{ page.lead | markdownify }}</p>
```
aduth, line 30: Hm, unfortunately it looks like `markdownify` will add its own wrapping `<p>` tag which is breaking the formatting from `usa-intro`. Maybe we could swap the wrapper with a `<div>` ?

```suggestion
            <div class="usa-intro">{{ page.lead | markdownify }}</div>
```

_layouts/documentation.html
```diff
@@ -43,7 +43,7 @@
         <div class="usa-display">{{ page.title }}</div>
 
         {% if page.lead %}
-          <p class="usa-intro">{{ page.lead }}</p>
+          <p class="usa-intro">{{ page.lead | markdownify }}</p>
```
aduth, line 46:
```suggestion
          <div class="usa-intro">{{ page.lead | markdownify }}</div>
```

aduth (approved):

Sgtpluck (approved):

PR #313, opened 2023-03-23
Add Jekyll plugin to apply design system styling to links

aduth:
Why: For consistent user experience across Login.gov properties.

This plugin is used already across a few Login.gov sites:

- [`18f/identity-style-guide`](https://github.com/18F/identity-style-guide/blob/main/docs/_plugins/content_typography.rb)
- [`18f/identity-handbook`](https://github.com/18F/identity-handbook/blob/main/_plugins/content_typography.rb)

Before|After
---|---
developers login gov_|localhost_54902_

### DIFF
--- a/_plugins/content_typography.rb
+++ b/_plugins/content_typography.rb
@@ -1,0 +1,12 @@
+module Kramdown
+  module Parser
+    class Kramdown
+      prepend(Module.new do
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

PR #331, opened 2023-05-02
Upgrade Login.gov Design System to v7.0.1

aduth:
## 🛠 Summary of changes

Upgrades the version of the Login.gov Design System from 6.7.0 to 7.0.1.

Primarily, this involves:

- Upgrading USWDS v2 to v3
- Migrating Sass syntax to new Dart-based `@use` & `@forward`

## 📜 Testing Plan

Preview link: https://federalist-14ccf384-8bd5-4e30-aa95-a5fe5894e3c1.sites.pages.cloud.gov/preview/18f/identity-dev-docs/aduth-lgds-7/

For changes introduced in the 7.0.1 release of the Login.gov Design System, it's worth special attention to navigation (both desktop and mobile), which received a significant refactor.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -13,7 +13,7 @@ whitelist:
   - jekyll-redirect-from
 
 copy_to_destination:
-  - node_modules/identity-style-guide/dist/assets
+  - node_modules/@18f/identity-design-system/dist/assets
 
 # Pages
 collections:
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -25,7 +25,6 @@
     <meta property="og:type" content="article">
 
     <link rel="shortcut icon" type="image/ico" href="{{ site.baseurl }}/assets/img/favicons/favicon.ico">
-    <link rel="stylesheet" href="{{ site.baseurl }}/assets/css/styles.css">
     <link rel="stylesheet" href="{{ site.baseurl }}/assets/css/main.css">
   </head>
 
@@ -142,7 +141,7 @@
             <div class="grid-col-auto">
               <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
             </div>
-            <div class="grid-col-auto tablet:grid-col-fill text-right">
+            <div class="grid-col-fill text-right">
               <a href="https://status.login.gov/">System Status</a>
             </div>
           </div>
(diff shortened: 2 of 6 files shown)


anniehirshman-gsa (approved):
This looks great - thank you! 

I had a couple questions about the new side navigation component that may or may not be relevant to the design system:

1. Can the side nav component be sticky/scrollable by default, or should this be configured at the site level?
2. Do the navigation levels need to be manually set/edited, or can they be determined automatically by the h2s and h3s on the page?
3. I noticed that the page nav now shifts to the right at tablet viewports (see image). I think this is a positive and we can fine-tune this at the site-level, but anything we should know/consider about why this occurs in v7.0.1 and not in previous versions?
4. FYSA: At mobile viewports, planning to implement a similar Menu solution as the brochure site [Help Center](https://federalist-17bd62cc-77b7-4687-9c62-39b462ce6fd5.sites.pages.cloud.gov/preview/18f/identity-site/aduth-lgds-7-0-1/help/) in upcoming work.

None of these should block merging in the upgrade - could be topics for a future LGDS discussion, or Team Ursula can address them in planned work on the page navigation. Thanks!

Screen Shot 2023-05-03 at 4 32 29 PM

zachmargolis (approved):

Sgtpluck (approved):
thank you!!!

aduth:
> 3\. I noticed that the page nav now shifts to the right at tablet viewports (see image). I think this is a positive and we can fine-tune this at the site-level, but anything we should know/consider about why this occurs in v7.0.1 and not in previous versions?

That's interesting and definitely not intended! I took a quick look and it seems like this is due to a behavior from USWDS which relies on the `desktop:` viewport range for its styling, and part of LGDS 7 was to remove that by default (only ship with "tablet"). I think what I'll do for the short-term is re-enable the desktop viewport utilities for this site, so that the navigation is unaffected.

But in follow-up work, I think we should reevaluate the navigation more holistically. At the very least, these USWDS classes (`usa-layout-docs__sidenav`) use `order` styling to shift around contents, which can have some very bad accessibility implications [[1]](https://developer.mozilla.org/en-US/docs/Web/CSS/order#accessibility_concerns) [[2]](https://github.com/uswds/uswds/issues/4424) [[3]](https://www.w3.org/TR/WCAG20-TECHS/C27.html).

anniehirshman-gsa:
Thanks @aduth for the additional info! Since I have a design ticket open to look at page nav more holistically on the developer guide, I'll consider different viewports and loop you in for feedback.

nickttng (approved):

PR #333, opened 2023-05-10
Update largest heading to be H1 instead of DIV

Jeremy1026:
### Relevant Ticket

[LG-9645](https://cm-jira.usa.gov/browse/LG-9645)

### Description of Changes

Updates `<div class="usa-display">` to be `<h1 class="usa-display">`, which updates the largest texts from divs to h1 for accessibility changes. Since the class remained the same, the UI has no changes.

image

image

image


aduth (approved):
LGTM 👍

PR #363, opened 2023-10-12
Sticky Left Nav

Jeremy1026:
### Relevant Ticket

[LG-9737](https://cm-jira.usa.gov/browse/LG-9737)

[Figma](https://www.figma.com/file/HHD28cToRlcOcqYcrFIm2o/Login.gov-Developer-Guide-(LG-9258%2C-LG-9244)?type=design&node-id=0-1&mode=design)

### Description of Changes

Adds the left nav inline to the top nav when in mobile.

### Screenshots

image

### DIFF
--- a/assets/scss/main.css.scss
+++ b/assets/scss/main.css.scss
@@ -32,3 +32,9 @@
 .reversefootnote {
   @include u-font-family("mono");
 }
+
+.usa-layout-docs__sidenav {
+  position: sticky;
+  top: 2rem;
+  height: 100vh;
+}
--- a/assets/scss/main.css.scss
+++ b/assets/scss/main.css.scss
@@ -34,7 +34,5 @@
 }
 
 .usa-layout-docs__sidenav {
-  position: sticky;
   top: 2rem;
-  height: 100vh;
 }
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -5,33 +5,6 @@ primary:
     href: /overview/
   - text: OpenID Connect
     href: /oidc/
-    sidenav:
-      - text: Getting started
-        href: "#getting-started"
-      - text: Authorization
-        href: "#authorization"
-        links:
-          - text: Authorization response
-            href: "#authorization-response"
-      - text: Token
-        href: "#token"
-        links:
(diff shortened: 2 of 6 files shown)


assets/scss/main.css.scss
```diff
@@ -32,3 +32,9 @@
 .reversefootnote {
   @include u-font-family("mono");
 }
+
+.usa-layout-docs__sidenav {
+  position: sticky;
+  top: 2rem;
+  height: 100vh;
```
nprimak, line 39: For setting the height we can use a utility class from uswds, `height-viewport`

https://designsystem.digital.gov/utilities/height-and-width/

assets/scss/main.css.scss
```diff
@@ -32,3 +32,9 @@
 .reversefootnote {
   @include u-font-family("mono");
 }
+
+.usa-layout-docs__sidenav {
+  position: sticky;
```
nprimak, line 37: For setting the position we can use a uswds utility class, `position-sticky`

https://designsystem.digital.gov/utilities/display/#position

nprimak (changes requested):
Since we're using USWDS for the dev docs and trying to keep custom css as minimal as possible, lets use the utility classes whenever possible

assets/scss/main.css.scss
Jeremy1026: Updated to use the USWDS class

assets/scss/main.css.scss
Jeremy1026: Updated to use the USWDS class

nprimak (approved):
Looking good!

anniehirshman-gsa (commented):
Mobile menu visually LGTM, thank you!

One comment: the sidebar is shown at the bottom of the page as well, which creates a big gap between the end of the main content and the footer/return to top/edit this page buttons. Could we hide the aside at mobile breakpoints at the bottom of the page, but keep it as-is in the mobile menu?

Screen Shot 2023-10-12 at 1 24 40 PM

anniehirshman-gsa (approved):
LGTM thank you!!

PR #370, opened 2023-10-20
Add word-wrap to stop having scroll in the X axis on mobile

Jeremy1026:
Add `word-wrap: break-word;` to `main.usa-layout-docs` to break long URLs that extend beyond the width of the device to stop the page being scrollable on the X axis.

Before:
image

After:
image


Sgtpluck (approved):
thanks for fixing!

nprimak (approved):

PR #377, opened 2023-11-01
LG-11480/add eslint

Sgtpluck:
Relevant ticket: https://cm-jira.usa.gov/browse/LG-11480

This change introduces eslint and prettier to the repo as well as circle-ci. It follows the [identity-site's](https://github.com/18F/identity-site/blob/d4e237a09aabcb23aacdc472c0a80f1df3053efe/package.json) pattern.

This repo currently only has one javascript file, which is being skipped because it is a copy/pasted minified file. However, https://cm-jira.usa.gov/browse/LG-10766 will be introducing javascript, so this PR is in preparation for that.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,10 +8,5 @@
     "parserOptions": {
       "ecmaVersion": 2021,
       "sourceType": "module"
-    },
-    "rules": {
-      // ESLint is currently unable to resolve packages with the exports directive
-      // See: https://github.com/import-js/eslint-plugin-import/issues/1810
-      "import/no-unresolved": ["error", { "ignore": ["@18f/identity-address-search", "@18f/identity-i18n", "@18f/identity-components"] }]
     }
   }
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,8 +3,7 @@
     "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
     "plugins": ["@18f/eslint-plugin-identity"],
     "env": {
-      "browser": true,
-      "jest": true
+      "browser": true
     },
     "parserOptions": {
       "ecmaVersion": 2021,
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,18 @@
+{
+    "root": true,
+    "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
+    "plugins": ["@18f/eslint-plugin-identity"],
+    "env": {
+      "browser": true,
+      "jest": true
+    },
+    "parserOptions": {
+      "ecmaVersion": 2021,
+      "sourceType": "module"
+    },
+    "rules": {
+      // ESLint is currently unable to resolve packages with the exports directive
+      // See: https://github.com/import-js/eslint-plugin-import/issues/1810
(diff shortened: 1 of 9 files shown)


.eslintrc
```diff
@@ -0,0 +1,18 @@
+{
+    "root": true,
+    "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
+    "plugins": ["@18f/eslint-plugin-identity"],
+    "env": {
+      "browser": true,
+      "jest": true
```
aduth, line 7: I don't think we'll need `jest` here since we're not using it (and frankly I hope we won't!)

```suggestion
      "browser": true
```

.eslintrc
```diff
@@ -0,0 +1,18 @@
...
+    },
+    "parserOptions": {
+      "ecmaVersion": 2021,
+      "sourceType": "module"
+    },
+    "rules": {
+      // ESLint is currently unable to resolve packages with the exports directive
+      // See: https://github.com/import-js/eslint-plugin-import/issues/1810
+      "import/no-unresolved": ["error", { "ignore": ["@18f/identity-address-search", "@18f/identity-i18n", "@18f/identity-components"] }]
+    }
```
aduth, line 17: Can we see if it works without these overrides? I think they were for some specific usage in `identity-site`.

```suggestion
    }
```

.prettierignore
```diff
@@ -0,0 +1,3 @@
+# Ignore artifacts:
+build
+coverage
```
aduth, line 3: Does this project generate these files?

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "singleQuote": true,
+  "trailingComma": "all",
```
aduth, line 3: This was recently changed to be the default in Prettier 3.0, so we can probably omit it.

```suggestion
```

See: https://prettier.io/blog/2023/07/05/3.0.0#javascript

.prettierignore
Sgtpluck: i am not sure! i poked around and couldn't tell, but i also wasn't sure how identity-site generates them, so i left this in to be safe. happy to remove it if i was just being overly cautious!

nprimak (approved):

.prettierignore
Sgtpluck: sounds good! love to remove things!!!

assets/js/tmp.js
aduth, line 1: What's the value in keeping this file until a "real" file exists? I think as long as we've verified it actually lints, we should trust that future files will be subjected to it.

.prettierrc.json
aduth, line 1: Do we plan to introduce Prettier dependencies and a lint step?

aduth (approved):

assets/js/tmp.js
Sgtpluck: because it throws an error and we can't merge without a real file to lint!

.prettierrc.json
Sgtpluck: the `@18f/eslint-plugin-identity` documentation says: `If eslint-plugin-prettier is installed, Prettier rules are enabled.` so i _think_ that since i included `eslint-plugin-prettier` in `package.json` means that the prettier config would run as part of eslint!

aduth: For posterity: @Sgtpluck and I synced in Slack, I forgot that the ESLint plugins will do the Prettier checks.

PR #375, opened 2023-10-31
LG-10766 - Authorization Page OIDC guide

nprimak:
https://cm-jira.usa.gov/browse/LG-10766

This is part of the dev doc redesign / rework with content updates and design updates broken down by page. This is for the Authorization page which has a few unique components as is a bit more complex design wise than the other pages.

When all page reworks are completed the navigation will be updated so that users can view the new pages. For now you must append /oidc/authorization to visit the page from home in Federalist preview. 

Biggest aspect of the rework is introduction of code snippets which scroll with the page and stay in the sidebar on the right hand side.

Figma can be seen here, please go to OIDC Guide section and Authorization page 
https://www.figma.com/file/3KSAr0qfDv8BXFIHyFEudZ/Login.gov-OIDC%2C-SAML%2C-and-UX-guides-(LG-10262%2C-LG-10514)?type=design&node-id=104-4001&mode=design

Note that the "return to top" button has been moved to the left so that it doesn't overlap with the gray code snippet sidebar and code snippets are hidden in mobile view.

PR #381, opened 2023-11-02
LG-11503 Getting started page for OIDC Guide rework

nprimak:
~~**This should not be merged until https://github.com/18F/identity-dev-docs/pull/375 is merged because it branched off of that branch.**~~ 
branch is now merged in 

This is part of the larger dev-doc epic for a redesign/overhaul of developer doc content. 

Please reference the Getting Started page under the OIDC section in Figma: https://www.figma.com/file/3KSAr0qfDv8BXFIHyFEudZ/Login.gov-OIDC%2C-SAML%2C-and-UX-guides-(LG-10262%2C-LG-10514)?type=design&node-id=104-4002&mode=design&t=0QdfxzCPUojzGdyr-0


_pages/oidc/getting-started.md
```diff
@@ -0,0 +1,86 @@
...
+  - text: Token
+    href: "oidc/#token"
+  - text: User info
+    href: "oidc/#user-info"
+  - text: Certificates
+    href: "oidc/#certificates"
+  - text: Logout
+    href: "oidc/#logout"
+  - text: Example application
+    href: "oidc/#example-application"
```
zachmargolis, line 30: is there any way to reuse this sidenav between pages? having to change 2+ other pages if a heading gets renamed seems like a lot of work

_pages/oidc/getting-started.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+## Getting started
+
+### Choosing an authentication method
+
+Login.gov supports two ways of authenticating clients: **private_key_jwt** and **PKCE**.
+
+- **private_key_jwt** (preferred for web apps)
+  The client sends a <a class="usa-link usa-link usa-link--external" href="https://jwt.io/">JSON Web Token</a>, or JWT, signed with a private key (minimum length of 2048 bits) when requesting access tokens. The corresponding public key is registered with the IdP ahead of time, similar to SAML.
```
zachmargolis, line 47:
```suggestion
  The client sends a <a class="usa-link usa-link--external" href="https://jwt.io/">JSON Web Token</a>, or JWT, signed with a private key (minimum length of 2048 bits) when requesting access tokens. The corresponding public key is registered with the IdP ahead of time, similar to SAML.
```
`usa-link` was added here twice. Can we do these as plain markdown syntax links and then maybe modify our `content_typography.rb` to detect external links and add the right class there?

https://github.com/18F/identity-dev-docs/blob/6f5568c261ffacafb9cd535232aa9de8ca6a16cf/_plugins/content_typography.rb#L17

_pages/oidc/getting-started.md
nprimak: hmm not sure... each page nav will be slightly different because the page you are on will have the sub sections on the current page 'expanded' and sub sections on other pages hidden

_pages/oidc/getting-started.md
nprimak: sure that page is almost all markdown so I would rather keep it as markdown wherever possible

nprimak: On second thought I know how I would do this with Javascript but not sure how to deduce what the host is using just plain ruby. In JS I would just check `window.host`

karenstrother (approved):
Looks good

_pages/oidc/getting-started.md
zachmargolis: We could see if the URL includes `site.baseurl` or not, if not, then it's probably external

zachmargolis: Alternatively, you could do `[link title](https://example.com){:class="usa-link--external"}` syntax

nprimak: done!

nprimak: @zachmargolis  mind taking another look ?

_pages/oidc/getting-started.md
```diff
@@ -0,0 +1,86 @@
...
+    href: "oidc/#logout"
+  - text: Example application
+    href: "oidc/#example-application"
+
+---
+<div class="usa-alert usa-alert--warning">
+<div class="usa-alert__body">
+<p class="usa-alert__text">We do not support the OpenID Connect “implicit flow” with client_secret because it is not recommended by the OAuth group for security reasons</p>
+</div>
+</div>
```
zachmargolis, line 37: We have the `alert` partial in this project so we ca replace this with
```suggestion
{% capture implicit_flow_warning %}
We do not support the OpenID Connect “implicit flow” with client_secret because it is not recommended by the OAuth group for security reasons
{% endcapture %}
{% include alert.html content=implicit_flow_warning alert_class="usa-alert--warning" %}
```

_pages/oidc/getting-started.md
```diff
@@ -0,0 +1,86 @@
...
+- [https://idp.int.identitysandbox.gov/.well-known/openid-configuration](https://idp.int.identitysandbox.gov/.well-known/openid-configuration)
+
+Production URI:
+- [https://secure.login.gov/.well-known/openid-configuration](https://secure.login.gov/.well-known/openid-configuration)
+
+### Example Application
+
+The Login.gov team has created an example client to speed up your development, all open source in the public domain: [identity-oidc-sinatra](https://github.com/18F/identity-oidc-sinatra){:class="usa-link--external"}.
+
+<a href="{{ site.baseurl }}/oidc/authorization/" class="usa-link margin-top-4">Next step: Authorization</a>
```
zachmargolis, line 86:
```suggestion
[Next step: Authorization]({{ site.baseurl }}/oidc/authorization/){:class="margin-top-4"}
```

zachmargolis (approved):
LGTM! one more suggestion to use markdown syntax, and one more spot we can use a shared partial

PR #382, opened 2023-11-03
LG 11484 Add error dictionary

Sgtpluck:
Relevant ticket:
https://cm-jira.usa.gov/browse/LG-11484

This change:
* Updates the `/support` view to have properly nested html. It removes unnecessary markdown references to make the HTML more readable.
* Uses a new YML file and the accordion component to create new  OIDC authorization request troubleshooting section

Notes: 
`invalid_verified_within_duration:` and `invalid_verified_within_format` are not being documented at the moment.

### DIFF
--- a/_data/errors.yml
+++ b/_data/errors.yml
@@ -4,52 +4,52 @@ oidc:
       id: bad-client-id
       accordion-id: bad-client
       content: |
-        ###### Why it's happening
-        The issuer (also known as the client id) provided does not match a service provider registered in the IdP
-        ###### What to do:
+        ##### Why it's happening
+        The issuer (also known as the client id) provided does not match a service provider registered in the IdP.
+        ##### What to do:
         - Ensure the authentication request has a `client_id` value
-        - Ensure that the `client_id` sent in the authentication request matches the one registered in your application on the [Partner Dashboard](https://dashboard.int.identitysandbox.gov/){:target="_blank"}
-        - Ensure that the application configuration's `active` value is set to `true`
+        - Ensure that the `client_id` sent in the authentication request matches the one registered in your application on the [Partner Dashboard](https://dashboard.int.identitysandbox.gov/){:target="_blank"}.
+        - Ensure that the application configuration's `active` value is set to `true`.
     - title: Missing a valid IAL level
       id: oidc-missing-ial
       accordion-id: oidc-missing-ial-level
       content: |
-        ###### Why it's happening
-        No IAL Service Levels are present in the `acr_values` parameter of the authentication request
-        ###### What to do:
+        ##### Why it's happening
+        No IAL Service Levels are present in the `acr_values` parameter of the authentication request.
+        ##### What to do:
         Check the `acr_values` parameter in your authentication request and ensure that you are passing either:
-        - `http://idmanagement.gov/ns/assurance/ial/1` for Authentication Only applications, OR
-        - `http://idmanagement.gov/ns/assurance/ial/2` for Identity-Verification Permitted applications
+        - `http://idmanagement.gov/ns/assurance/ial/1` for Authentication Only applications, OR;
+        - `http://idmanagement.gov/ns/assurance/ial/2` for Identity-Verification Permitted applications.
         Learn more about IAL values here: [https://developers.login.gov/oidc/#ial-values](https://developers.login.gov/oidc/#ial-values){:target="_blank"}
     - title: The acr_values are not authorized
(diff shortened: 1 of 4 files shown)


Sgtpluck:
Screenshot 2023-11-03 at 4 16 18 PM

Screenshot 2023-11-03 at 4 16 31 PM

_data/errors.yml
```diff
@@ -0,0 +1,93 @@
+oidc:
+  auth:
+    - title: Bad client_id
+      id: bad-client-id
+      accordion-id: bad-client
+      content: |
+        ###### Why it's happening
```
nprimak, line 7: Can we use h5's on all of these instead of h6's? They are bigger and read more like headers to me visually than h6 which is a little hard to read.

nprimak (approved):
Thank you for fixing the html on the support page! I just left one comment regarding the headers

_data/errors.yml
aduth: This is also important for accessibility, as it looks like H5 would be the correct next level in the heading order.

https://dequeuniversity.com/rules/axe/3.5/heading-order

It'd be good to get some accessibility testing in place since it looks like there's already some existing issues related to this.

_data/errors.yml
```diff
@@ -0,0 +1,93 @@
...
+        The issuer (also known as the client id) provided does not match a service provider registered in the IdP
```
karenstrother, line 8: The "why its happening" sentences need periods.

_data/errors.yml
```diff
@@ -0,0 +1,93 @@
...
+        ###### What to do:
+        - Ensure the authentication request has a `client_id` value
```
karenstrother, line 10: These instructions in the "what to do" sections also need periods.

_data/errors.yml
```diff
@@ -0,0 +1,93 @@
...
+        - If this is a production instance, submit a [Zendesk request](https://zendesk.login.gov){:target="_blank"} to deploy that change to production.
+    - title: No acceptable acr_values found
+      id: oidc-no-acr-values
+      accordion-id: no-acr
+      content: |
+        ###### Why it's happening
+        No AAL levels are present in the acr_values parameter of the authentication request
+        ###### What to do:
+        - Check the `acr_values` parameter in your authentication request.
+        = Ensure that you are passing in the appropriate IAL value as well as any AAL values you might want for your application.
```
karenstrother, line 44: Is the = a typo?

_data/errors.yml
```diff
@@ -0,0 +1,93 @@
...
+        ###### What to do:
+        - Check the `scope` parameter in your authentication request.
+        - Ensure you are passing in `openid` as well as at least one other attribute.
+        More information about the different types of attributes can be found here: [https://developers.login.gov/attributes/](https://developers.login.gov/attributes){:target="_blank"}
+    - title: No valid prompt values found
+      id: oidc-no-prompt
+      accordion-id: no-prompt
+      content: |
+        ###### Why it's happening
+        No prompt attribute was passed through the authentication request
```
karenstrother, line 61: If still technically correct, I suggest changing this sentence to match the one above - "No prompt attribute was passed in the `prompt` parameter of the authentication request." If that is no longer correct please disregard.

_data/errors.yml
Sgtpluck: ah yes sorry, i think this was a typo that got copy/paste propagated, apologies. 

agree with the accessibility testing, although i think that might have to be its own PR depending on how many issues crop up. is there a preferred automated testing tool for login.gov?

Sgtpluck: fixed the headers in https://github.com/18F/identity-dev-docs/pull/382/commits/ec23adf4753192c2c0a61891f6594a25ae604f70

_data/errors.yml
Sgtpluck: fixed in https://github.com/18F/identity-dev-docs/pull/382/commits/ec23adf4753192c2c0a61891f6594a25ae604f70

_data/errors.yml
Sgtpluck: fixed in https://github.com/18F/identity-dev-docs/pull/382/commits/ec23adf4753192c2c0a61891f6594a25ae604f70

_data/errors.yml
Sgtpluck: fixed in https://github.com/18F/identity-dev-docs/pull/382/commits/ec23adf4753192c2c0a61891f6594a25ae604f70

PR #387, opened 2023-11-09
LG-11413 developer doc logout page rework

nprimak:
Part of the ongoing epic reworking the developer docs

This is specifically for the **logout** page, please compare to mockup here in Figma:
https://www.figma.com/file/3KSAr0qfDv8BXFIHyFEudZ/Login.gov-OIDC%2C-SAML%2C-and-UX-guides-(LG-10262%2C-LG-10514)?type=design&node-id=104-4001&mode=design&t=YDYdtdA55wzFHwf2-0

~This is the last page for the OIDC section but will not be visible to users until the SAML guide is also complete.~
This PR will make the OIDC pages live in production


Sgtpluck (commented):
One small thing I noticed is the logout url needs to be updated in `certificates.md` it is currently still linking back to the original view with the `#logout` syntax

Sgtpluck (approved):

assets/js/toggle_view.js
```diff
@@ -87,3 +89,7 @@ if (oidcUserInfoTab1Button) {
   oidcUserInfoTab1Button.addEventListener('click', showTabOneCode);
   oidcUserInfoTab2Button.addEventListener('click', showTabTwoCode);
 }
+if (oidcLogoutTab1Button) {
+  oidcLogoutTab1Button.addEventListener('click', showTabOneCode);
+  oidcLogoutTab2Button.addEventListener('click', showTabTwoCode);
+}
```
aduth, line 95: Is there any way we could generalize this so we don't need to have 4 different sets of conditions? e.g. if we applied a common class to the tab buttons so that we add an event listener to whichever element has that class, regardless which one it is?

PR #402, opened 2023-11-29
LG-11645: production page updates

nprimak:
https://cm-jira.usa.gov/browse/LG-11645

Updated the markdown to match other pages (use hashtags for headers, and alert component), tweak the bolded sections to make it easier to read, and add some additional text to staging environments section. 

Because this section had lead text, I decided to keep the blue hero (note in Figma the lead text/title  is not in a hero). I am open to changing this if anyone has strong feelings. 

Figma link: https://www.figma.com/file/3KSAr0qfDv8BXFIHyFEudZ/Login.gov-OIDC%2C-SAML%2C-and-UX-guides-(LG-10262%2C-LG-10514)?type=design&node-id=104-4001&mode=design&t=f1gOTHwL1FJl9Vjg-0

### DIFF
--- a/_pages/production.md
+++ b/_pages/production.md
@@ -36,8 +36,8 @@ Make sure you have the following items ready before you start the deployment pro
 
 -   [Signed Interagency Agreement (IAA) listing this integration ](https://developers.login.gov/production/#confirm-iaa)
 
--   A new dedicated [application configuration within the Dashboard](https://dashboard.int.identitysandbox.gov/)
-    * Some partners choose to have two configurations, one that is intended for deployment to production and one which is purely for testing purposes.
+-   A dedicated [application configuration within the Dashboard](https://dashboard.int.identitysandbox.gov/)
+    * We recommend having two configurations, one that is intended for deployment to production and one which is purely for testing purposes.
     * All production urls should have .gov, .mil, or a dedicated .com address and point to an Authority to Operate (ATO) approved environment.
 
 -   A user account in the [Login.gov production environment](https://secure.login.gov)
@@ -88,7 +88,7 @@ Before you can request deployment, you need to create a new and separate applica
 
 4.  Enter the name of the app as it should appear to users in the "Friendly Name" field. **This is the app name a user will see when logging in.** 
 
-5.  Next, choose an agency team for the app from the drop down menu
+5.  Next, choose an agency team for the app from the drop down menu.
 
 6.  Select the authentication protocol that will be used by the service provider. 
 
--- a/_pages/production.md
+++ b/_pages/production.md
@@ -82,25 +82,27 @@ Before you can request deployment, you need to create a new and separate applica
 
 1.  [Create a new app on the](https://dashboard.int.identitysandbox.gov/) Login.gov Partner Dashboard. Select “Apps” from the top right menu, then select the “Create a new test app” button.
 
-2.  Enter the name of the app as it appears in your IAA in the "App Name" field. **Do NOT use environment names such as "Prod" or "Production".**
+2. Select “YES” for production configuration - this configuration is for a production app.
 
-3.  Enter the name of the app as it should appear to users in the "Friendly Name" field. **This is the app name a user will see when logging in.** 
+3.  Enter the name of the app as it appears in your IAA in the "App Name" field. **Do NOT use environment names such as "Prod" or "Production".**
 
-4.  Next, choose an agency team for the app from the drop down menu
+4.  Enter the name of the app as it should appear to users in the "Friendly Name" field. **This is the app name a user will see when logging in.** 
 
-5.  Select the authentication protocol that will be used by the service provider. 
+5.  Next, choose an agency team for the app from the drop down menu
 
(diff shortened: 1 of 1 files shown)


Jeremy1026 (approved):

_pages/production.md
```diff
@@ -24,20 +24,20 @@ sidenav:
     href: "#certificate-rotation-process"
 saml_year: 2023
 ---
-<div class="usa-alert usa-alert--warning">
-<div class="usa-alert__body">
-<p class="usa-alert__text"><b> Please note: The deployment process can take up to two weeks to complete. You may wish to have a grace period between when we deploy and when you add the Login.gov button to your site. </b></p>
-</div>
-</div>
+{% capture deployment_warning %}
+Please note: The deployment process can take up to two weeks to complete. You may wish to have a grace period between when we deploy and when you add the Login.gov button to your site.
```
karenstrother, line 28: This was bolded previously.

_pages/production.md
```diff
@@ -46,28 +46,20 @@ Make sure you have the following items ready before you start the deployment pro
...
-
--   If this is a SAML integration (Not OpenID Connect), then please ensure that:
-
-    -   Assertion Consumer Logout Service URL is defined.
+- **If this is an integration requesting identity proofed attributes, you must include a Failure to Proof URL.** Users will be redirected to this URL if they fail to complete the identity verification process. This page should communicate your agency and/or departments alternate methods of accessing your application.
 
-    -   SAML Assertion Encryption is enabled.
+If you have questions after reviewing this page, submit a technical support ticket through the [Partner Support Help Desk]({{ site.baseurl}}/support/#contacting-partner-support). You will need a [Login.gov production account](https://secure.login.gov) to submit technical support tickets. Your Login.gov production account and Login.gov sandbox (test environment) accounts are separate.
```
karenstrother, line 59: flagging Sandbox here - but this is a case where I am not sure replacing sandbox with dashboard is entirely accurate.

karenstrother (approved):

_pages/production.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Production deployment
 lead: >
-  Once you’ve tested your integration in [our sandbox environment](https://dashboard.int.identitysandbox.gov/), you can request deployment to the Login.gov production environment.
+  Once you’ve setup and tested your integration in [our dashboard](https://dashboard.int.identitysandbox.gov/), you can request deployment to the Login.gov production environment.
```
Sgtpluck, line 4: [nitpick] `set up` (setup is a noun, set up is a verb)

Sgtpluck: question -- is this confusing? The partner sets up their configuration in the partner dashboard, and then tests in their sandbox evironment. this feels like it implies that there is testing that is possible in the dashboard. (i know this isn't really a content change ticket, though)

_pages/production.md
```diff
@@ -24,20 +24,20 @@ sidenav:
...
+{% include alert.html content=deployment_warning alert_class="usa-alert--warning" %}
+
+## Before deployment
 
 Make sure you have the following items ready before you start the deployment process:
 
 -   [Signed Interagency Agreement (IAA) listing this integration ](https://developers.login.gov/production/#confirm-iaa)
 
--   A new dedicated [production app configuration within the Dashboard](https://dashboard.int.identitysandbox.gov/)
+-   A new dedicated [application configuration within the Dashboard](https://dashboard.int.identitysandbox.gov/)
```
Sgtpluck, line 39: [nitpick] technically it doesn't need to be "new". what do you think about dropping that word?

_pages/production.md
```diff
@@ -24,20 +24,20 @@ sidenav:
...
+    * Some partners choose to have two configurations, one that is intended for deployment to production and one which is purely for testing purposes.
```
Sgtpluck, line 40: should we state this as a recommendation rather than just an option?

_pages/production.md
Sgtpluck: sandbox is correct here @karenstrother ! (it is differentiating between login.gov accounts.) arguably we should call it the integration environment rather than test environment? (but we're not very consistent about that syntax.)

Sgtpluck (approved):
a couple of nitpicks/questions, but it looks good! thanks for your work!

_pages/production.md
nprimak: I think since it is already highlighted as a warning having it bolded is a bit excessive? None of the warnings on other pages are bolded that I have seen

_pages/production.md
nprimak: yeah I was on the fence about that myself - lets drop it 👍🏻

_pages/production.md
nprimak: makes sense to me!

_pages/production.md
nprimak: yeah I agree with Davida's comment below that in this case we are describing the accounts so I don't think we want to change the language here

_pages/production.md
karenstrother: What about "Once you've set up your integration through our dashboard and tested..."? Or I'd say leave it for now and we deal with the sandbox problem across the site all at once. Somehow.

PR #423, opened 2024-01-08
Move UX warning up to make it more visible

Sgtpluck:
Ticket:
https://cm-jira.usa.gov/browse/LG-10824

Cleans up Logout UX-impact text and moves it up the page

Section of logout doc text with the user experience impact described

Question: do we need similar text for the SAML logout view? Currently we don't have a similar warning there


nprimak (approved):
Approved! -  thinking ahead would also be nice to have an image showing what the user experience looks like

PR #425, opened 2024-01-10
Update documentation and yml files for selfie

night-jellyfish:
This change:

- updates the `sample_full_error.yml` for selfie entries, with the possible values notated
- updates the text in the yml section to include selfie
- removes a note that I don't think is true anymore (`Even if you put all passing information into the test yaml file it will still produce an error.` - you can upload a yml file that passes doc auth)
- updates the yml docs to include a note that only the back image is evaluated 
- updates the general documentation for a note about changing yml files since I had trouble finding a note about it

Note: These changes should not be merged until https://github.com/18F/identity-idp/pull/9882 is merged.

### DIFF
--- a/_pages/testing.md
+++ b/_pages/testing.md
@@ -141,7 +141,6 @@ Alert names with attention or failed values show under `failed_alerts`. Only pas
 **NOTES:**
 
 - There are configurations of the above yaml file that cannot happen in real vendor responses. It is possible there will be unexpected outcomes in those cases.
-- When using the `yaml` files for local dev, only the yml file supplied to the `back` is evaluated.
 
 ### Personal information verification
 
--- a/_pages/testing.md
+++ b/_pages/testing.md
@@ -93,7 +93,9 @@ Login.gov prompts users to upload the front and back of their documents during p
 
 #### Data testing
 
-A YAML file can be uploaded instead of a State ID image to trigger different behaviors. You will upload this text file for the front and back for the State ID, and possibly the selfie. The YAML file can be used to simulate the reading of certain attributes from the State ID or selfie. Here is an example YAML file that does that:
+⚠️ Only the YAML file in the `back` upload box is used on submission. For that reason, it's good practice to upload the same yaml file in all three upload boxes to avoid confusion when testing.
+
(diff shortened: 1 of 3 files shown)


_pages/testing.md
```diff
@@ -93,7 +93,7 @@ Login.gov prompts users to upload the front and back of their documents during p
 
 #### Data testing
 
-A YAML file can be uploaded instead of a State ID image to trigger different behaviors. You will upload this text file for the front and back for the State ID. The YAML file can be used to simulate the reading of certain attributes from the State ID. Here is an example YAML file that does that:
+A YAML file can be uploaded instead of a State ID image to trigger different behaviors. You will upload this text file for the front and back for the State ID, and possibly the selfie. The YAML file can be used to simulate the reading of certain attributes from the State ID or selfie. Here is an example YAML file that does that:
```
charleyf, line 96:
```suggestion
⚠️ Only the YAML file in the `back` upload box is used on submission. For that reason, it's good practice to upload the same yaml file in all three upload boxes to avoid confusion when testing.

A YAML file can be uploaded instead of a State ID image to trigger different behaviors. You will upload the same YAML file for the front and back of the State ID, and also the selfie (if that field appears). Different YAML files can be used to simulate the reading of certain attributes from the State ID or selfie. Here is an example YAML file that does that:
```

night-jellyfish: Thank you @charleyf! I had a similar note further down, but I think this is more clear.

_pages/testing.md
```diff
@@ -136,7 +138,10 @@ Alert names with attention or failed values show under `failed_alerts`. Only pas
       - name: Visible Photo Characteristics
 ```
 
-**NOTE:** Even if you put all passing information into the test yaml file it will still produce an error. There are configurations of the above yaml file that cannot happen in real vendor responses. It is possible there will be unexpected outcomes in those cases.
+**NOTES:**
+
+- There are configurations of the above yaml file that cannot happen in real vendor responses. It is possible there will be unexpected outcomes in those cases.
+- When using the `yaml` files for local dev, only the yml file supplied to the `back` is evaluated.
```
night-jellyfish, line 144:
```suggestion
```

night-jellyfish:
https://github.com/18F/identity-idp/pull/9882 has now been merged so this can be merged when ready.

charleyf (approved):
Text changes look good to me.

PR #443, opened 2024-01-22
Move from RSpec internal validity page check to html-proofer (LG-10193)

mitchellhenke:
[LG-10193](https://cm-jira.usa.gov/browse/LG-10193)
[LG-12164](https://cm-jira.usa.gov/browse/LG-12164)

Similar to what we did in https://github.com/18F/identity-site/pull/658 for the brochure site, this PR removes the internal custom RSpec matcher and replaces it with html-proofer.

Many of the changes are converting paths prepend the jekyll baseurl and fixing anchors in hrefs. Going forward, all links will have to be valid if this is merged.


_layouts/base.html
```diff
@@ -95,7 +95,7 @@
       <header class="usa-header usa-header--extended" role="banner">
           <div class="usa-navbar">
             <div class="usa-logo" id="basic-logo">
-              <a href="{{ site.baseurl }}/" title="Home" aria-label="Home">
```
aduth, line 98: Wouldn't we want to keep `site.baseurl` as much as possible to support links in Federalist previews resolving correctly?

.circleci/config.yml
```diff
@@ -33,7 +33,9 @@ jobs:
           command: |
             make clean
             make test
-
+      - run:
+          name: Run htmlproofer (no external links)
+          command: bundle exec scripts/htmlproofer --internal
```
mitchellhenke, line 38: The `--internal` flag means we will only check internal links. We've found external ones can be a little flaky.

_layouts/base.html
mitchellhenke: yep. we had mixed usage previously and I saw that `identity-site` wasn't using `site.baseurl` (but those appear to just be a few stragglers). Added 6e862e3 to fix it.

zachmargolis (approved):

Sgtpluck (approved):
thank you so much!

PR #451, opened 2024-02-15
Updates documentation for new SAML year

Sgtpluck:
A [partner pointed out](https://logingov.zendesk.com/agent/tickets/7826) that our SAML signing certificate documentation was out of date. 

This change:
- Removes reference to the staging certificate
- Updates the certificates to the 2024 certificates
- Adds a site-wide variable for the saml year, rather than it being added ad-hoc to different pages
- Updates the saml year


_data/saml.yml
```diff
@@ -0,0 +1,3 @@
+year:
+  current: 2024
+  previous: 2023
```
zachmargolis, line 3: love this! would it be worth putting the SAML certs in here too so there's only one file to touch every year?

zachmargolis (approved):

_data/saml.yml
Sgtpluck: oh yeah i think that will make it super easy

zachmargolis: big fan big fan

PR #457, opened 2024-02-22
Add CSS to fix alert icons

Jeremy1026:
This Pull Request fixes the icons in the warning boxes. Currently the CSS on the `before` pseudo element has the styles 

```
height: 1rem;
width: 1rem;
content: "";
height: 2px;
background-image: url(../img/usa-icons/warning.svg);
background-position: 50%;
background-repeat: no-repeat;
background-size: 1rem 1rem;
display: block;
position: absolute;
top: .5625rem;
left: 1rem;
```

With `height: 2px` taking precedent over the `height: 1rem`. I'm re-adding the `height: 1rem` with higher precedent to "win."

I know we don't like to add CSS directly, but I can not find where the `height: 2px` is being added to the element's style.

Before:

image

After:

image

### DIFF
--- a/assets/scss/main.css.scss
+++ b/assets/scss/main.css.scss
@@ -146,12 +146,6 @@ button.code-button:hover {
   width: 230px;
 }
 
-.usa-alert {
-  .usa-alert__body::before {
-    height: 1rem;
-  }
-}
-
 @media (min-width: 1024px) {
   .code-snippet-column {
     background-color: transparent;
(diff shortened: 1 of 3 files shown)


aduth:
This [bug](https://github.com/uswds/uswds/pull/5187) is fixed in newer versions of the design system, can we update instead?

Specifically:

https://github.com/18F/identity-dev-docs/blob/bc0b6e051893308452a0bfafbec596474604c084/package.json#L22

Upgrading to at least `@18f/identity-design-system@8.0.0` should fix this.

Jeremy1026:
Cancelling this PR because updating the design system _and_ build-sass worked!

PR #467, opened 2024-03-25
LG-12616: Add axe accessibility scan

nprimak:
[LG-12616](https://cm-jira.usa.gov/browse/LG-12616)

Setting up Axe / Puppeteer to run tests - thank you to @aduth  borrowing from [similar implementation](https://github.com/18F/identity-design-system/pull/435) done last week on identity-design-system.

~~After running the tests there are failures with missing `title` element and `lang` attribute on practically all the pages so I have commented out the line in the Makefile until those failures are addressed.~~

There are quite a few errors on a lot of different pages that need to be addresses ranging from critical to mild. 
 See full log output in [Google Doc here](https://docs.google.com/document/d/1Ihfi6tINgbrw9PTcgsQfuWk-UQ-ndJVZALL8EbWTPGw/edit#heading=h.94dqpr3pizkw).

The plan is to  [address the failures in this ticket](https://cm-jira.usa.gov/browse/LG-12800) and then we can update the `disableRules` method on line 50 appropriately as we address the failures.


Jeremy1026 (dismissed):

package.json
```diff
@@ -19,13 +20,18 @@
   "homepage": "https://github.com/GSA-TTS/identity-dev-docs#readme",
   "dependencies": {
     "@18f/identity-build-sass": "^3.0.0",
-    "@18f/identity-design-system": "^8.1.2"
+    "@18f/identity-design-system": "^8.1.2",
+    "puppeteer": "^22.6.1"
```
aduth, line 24: I think this should be moved to `devDependencies` for consistency with the other dependencies and since it wouldn't be required as part of the standard build of the site.

spec/e2e/accessibility.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('has pages to test', () => {
+    assert(paths.length);
+  });
+
+  paths.forEach((path) => {
+    test(path, async () => {
+      const page = await browser.newPage();
+      await page.goto(`http://localhost:${port}${path}`);
+      const results = await new AxePuppeteer(page).withTags(['wcag2a', 'wcag2aa']).analyze();
```
aduth, line 52: Rather than not running the tests at all, I'd suggest disabling the rules that will be addressed in the follow-on ticket, so that we could at least avoid any _new_ issues being introduced.

```suggestion
      const results = await new AxePuppeteer(page)
        .withTags(['wcag2a', 'wcag2aa'])
        // TODO: These will be addressed in LG-12800
        .disableRules(['document-title', 'html-has-lang', 'meta-viewport'])
        .analyze();
```

Makefile
```diff
@@ -12,6 +12,8 @@ clean:
 
 test: build
 	bundle exec rspec spec
+	# must fix existing accessibility errors before uncommenting e2e tests
+	# node --test spec/e2e/
```
aduth, line 16: Related to previous comment:

```suggestion
	node --test spec/e2e/
```

spec/e2e/accessibility.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+  /** @type {import('puppeteer').Browser} */
+  let browser;
+
+  before(async () => {
+    publicPath = sitepath.relative(process.cwd(), sitepath.resolve('../../_site'));
+    port = await getPort();
+    browser = await puppeteer.launch();
+    server = createServer((request, response) =>
+      handler(request, response, { public: publicPath }),
```
aduth, line 36: I'm trying to follow the logic for `publicPath` and wondering if it's even necessary to have at all, vs. just assuming the tests are run from the root directory.

```suggestion
      handler(request, response, { public: '_site' }),
```

aduth (approved):
LGTM 👍

Sgtpluck (approved):

spec/e2e/accessibility.test.js
```diff
@@ -0,0 +1,66 @@
...
+  it('has pages to test', () => {
+    assert(paths.length);
+  });
+
+  paths.forEach((path) => {
+    test(path, async () => {
+      const page = await browser.newPage();
+      await page.goto(`http://localhost:${port}${path}`);
+      const results = await new AxePuppeteer(page)
+        .withTags(['wcag2a', 'wcag2aa'])
```
aduth, line 48: Depending how strict you want the checks, there's a number of "best practices" rules that aren't strictly tied to WCAG that you could opt-in to. I suspect it'd catch a handful more issues, but issues still worth taking a look at. Up to y'all 🤷 

Related: https://github.com/18F/identity-design-system/pull/435#discussion_r1532727802

.circleci/config.yml
```diff
@@ -5,11 +5,24 @@ jobs:
...
+      - image: cimg/ruby:3.1-browsers
 
     working_directory: ~/identity-dev-docs
     steps:
       - checkout
+      
+      - run:
+          name: Switch Node.js version
+          command: |
+            wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
```
ajfarkas, line 17: out of curiosity, why `0.39.0` and not `0.39.7`? Is there a better way to install `^0.39`?

ajfarkas: hm, this version is from 10/2021

ajfarkas (changes requested):

spec/e2e/accessibility.test.js
Sgtpluck: i think having stricter checks would be good! we should fix the best practice issues.

.circleci/config.yml
aduth: That's a good call-out. This was [copied from `identity-site`](https://github.com/GSA-TTS/identity-site/blob/9a91f4fed83f8818843244913e4ea87e77af1efb/.circleci/config.yml#L11-L18), where it was the latest version at the [time of introduction](https://github.com/GSA-TTS/identity-site/pull/990). It was pinned to the specific version to avoid unexpected breakage from changing versions and for security, though in retrospect there's not much security assurances relying on git tags, since they can be deleted and recreated. SHA might be a better bet.

I'd think this could be bumped to the current latest version, assuming it works.

Jeremy1026 (approved):

.circleci/config.yml
nprimak: I just grabbed the same version/command that the identity-design-system is using for the sake of consistency -- https://github.com/18F/identity-design-system/blob/948b3326379f182fd4d5dbabcc10daca510763d7/.circleci/config.yml#L13-L20

spec/e2e/accessibility.test.js
nprimak: Thanks for flagging this. I will include it in the follow up ticket so we can add them once we've knocked out some of the higher urgency items.

PR #476, opened 2024-04-17
Update sample PII in test documents to match results of `pii_from_doc` in the IdP

jmhooper:
The `DocAuthMockClient` returns the value under the `document` key in the YAML test values ([ref](https://github.com/18F/identity-idp/blob/c53dce8f6679d25d29ba9a7d71911a58c0748f12/app/services/doc_auth/mock/result_response.rb#L83)).

This commit changes the value under this key in the sample YAML documents to match what is returned by the actual doc auth client. That was done by referencing the `DocAuth::LexisNexis::DocPiiReader` which is used to read the PII from the response and construct the `pii_from_doc` response for the TrueID client ([ref](https://github.com/18F/identity-idp/blob/c53dce8f6679d25d29ba9a7d71911a58c0748f12/app/services/doc_auth/lexis_nexis/doc_pii_reader.rb)).

### DIFF
--- a/assets/yaml/individual_state_error.yml
+++ b/assets/yaml/individual_state_error.yml
@@ -1,5 +1,4 @@
 document:
-  type: license
   first_name: Susan
   last_name: Smith
   middle_name: Q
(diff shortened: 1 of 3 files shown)


amirbey (approved):
LGTM

PR #477, opened 2024-04-25
Maintenance updates

monfresh:
This PR contains some maintenance updates:

- Update Ruby to 3.2.4
- Update jekyll, nokogiri, rspec, bundler to latest versions
- Fix warning about excerpt when running local server
- Update CircleCI config.yml to take advantage of orbs

See the individual commit messages for more details.


Sgtpluck (approved):
thank you!!! 🙏

ajfarkas (approved):
This is great to see.
Are we not using `store_test_results` or `store_artifacts` in the circleCI config?

.circleci/config.yml
```diff
@@ -1,61 +1,39 @@
...
-      - run:
-          name: Switch Node.js version
-          command: |
-            wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
-            export NVM_DIR="$HOME/.nvm"
-            . "$NVM_DIR/nvm.sh" --install --latest-npm
-            echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV;
-            echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV;
-      
+      - node/install
```
aduth, line 17: Neat, I didn't know CircleCI had [built-in support for NVM](https://circleci.com/developer/orbs/orb/circleci/node#commands-install)! That feels a lot nicer / safer than running code directly from GitHub 😅 I wonder if that's always been a thing, or if it's a new feature.

.circleci/config.yml
```diff
@@ -1,61 +1,39 @@
...
-      - save_cache:
-          key: v2-identity-dev-docs-{{ checksum "package-lock.json" }}-{{ checksum "Gemfile.lock" }}
-          paths:
-            - vendor/bundle
-            - ~/.npm
-
+      - checkout
+      - browser-tools/install-chromedriver
+      - node/install-packages
+      - ruby/install-deps
```
aduth, line 24: Also didn't realize they had these conveniences with built-in caching 🎉

.circleci/config.yml
monfresh: I think this feature is relatively new. I forget when I first discovered it. Maybe in the past 2 years? It looks like the Node Orb was originally created in 2018, but I'm not sure if it had all of today's functionality back then.

.circleci/config.yml
monfresh: Yeah, and built-in test reporting as well.

monfresh:
@ajfarkas That's a good question. Previously, the `store_test_results` and `store_artifacts` steps weren't doing anything because the config never created the `/tmp/test-results` folder 😄 You can verify this by looking at this old workflow, and see the error under the "Uploading Test Results" section: https://app.circleci.com/pipelines/github/GSA-TTS/identity-dev-docs/1131/workflows/35c7197a-c963-4381-8ed1-652aee33e6a4/jobs/1363

With this new config, RSpec test results should be uploaded automatically, but it looks like the Node tests are not. Either way, given the nature of the tests, and that they run in about 30 seconds, I don't think we have much use for CircleCI's Test Insights.

PR #478, opened 2024-04-26
LG-13067 Add official Login.gov Footer

ajfarkas:
[LG-13067 Add official Login.gov Footer](https://cm-jira.usa.gov/browse/LG-13067)

Because we don't have any localizations other than en-US, I made this pretty straightforward. If we do translations, we'll have to do some file organization, but I felt that was out of scope for this ticket.

[Figma design](https://www.figma.com/file/0AZ8vLjbLZYiIe1oteOt2I/developer.login.gov-footer?type=design&node-id=2-66&mode=design&t=keRhUB4TPZl1G4l4-0)


_includes/footer.html
```diff
@@ -0,0 +1,80 @@
...
+        <div class="usa-identifier__logos">
+          <a href="https://www.gsa.gov" class="usa-identifier__logo">
+            <img class="usa-identifier__logo-img" src="{{ site.baseurl }}/assets/img/gsa-logo.svg"
+              alt="{{ site.data.settings.identifier.accessible_labels.gsa_link_logo }}" />
+          </a>
+        </div>
+        <div class="usa-identifier__identity" aria-label="{{ site.data.settings.identifier.accessible_labels.description }}">
+          <p class="usa-identifier__identity-domain">Login.gov</p>
+          <p class="usa-identifier__identity-disclaimer">
+            {{ site.data.settings.identifier.disclaimer | markdownify | replace: '<p>', '' | replace: '</p>', '' }}
```
Sgtpluck, line 16: [question] Are the replacements necessary? i don't see any p tags in the yml

_data/settings.yml
```diff
@@ -0,0 +1,23 @@
+identifier:
+  accessible_labels:
+    description: Agency description
+    gsa_link_logo: GSA homepage
+    identifier: Agency identifier
+    links: Important links
+    login_gov_logo: Login.gov logo
+    usa_gov: U.S. government information and services
+  disclaimer: An official website of the [General Services
```
Sgtpluck, line 9: [thought] this line break is is a bit jarring. did the linter do this? i wasn't aware that we really enforced line limits (especially ones this short)

_includes/footer-navigation.html
```diff
@@ -0,0 +1,21 @@
+{% assign logo = include.logo | default: 'logo.svg' %}
```
Sgtpluck, line 1: [suggestion] it looks like we aren't passing anything in the one place we're rendering this partial. we should probably just directly set these values until we need to dynamically replace them

_data/nav.yml
```diff
@@ -19,3 +19,15 @@ primary:
...
+    become_a_partner: '$BASE_URL/partners/'
+    developer_guide: 'https://developers.login.gov'
+  learn:
+    about_us: '$LOCALE_BASE_URL/about-us/'
+    accessibility: '$LOCALE_BASE_URL/accessibility/'
+    join_us: '$LOCALE_BASE_URL/join/'
+    policies: '$LOCALE_BASE_URL/policy/'
+  support:
+    contact_us: '$LOCALE_BASE_URL/contact/'
+    help_center: '$LOCALE_BASE_URL/help/'
```
zachmargolis, line 33: does this variable work? I'd expect that we need to do like `| prepend: site.baseurl` in the templates

assets/scss/_variables.scss
```diff
@@ -0,0 +1,26 @@
+@use 'uswds-core' as *;
+@use 'colors' as *;
+
+$base-font-size: 1rem;
```
Sgtpluck, line 4: [question] i'm not seeing the majority of these variables being invoked. are they necessary to add?

_data/nav.yml
aduth: I think this is based on how it's implemented in `identity-site`, but it requires an additional piece to replace the variable fragments:

https://github.com/GSA-TTS/identity-site/blob/ab27dfeb90459fddabe5258d32f90cb3b6adc3df/_includes/components/footer-navigation.html#L34

Since we don't need localization, maybe we can just drop the  "LOCALE_BASE_URL" bits?

_data/settings.yml
aduth: I don't think there's a lint for this currently. That being said, we _could_ have one using the published [`@18f/identity-normalize-yaml`](https://github.com/18F/identity-idp/tree/main/app/javascript/packages/normalize-yaml), which respects the local Prettier configuration (including [print width](https://prettier.io/docs/en/options.html#print-width)) and which we use in other static sites [like `identity-site`](https://github.com/GSA-TTS/identity-site/blob/ab27dfeb90459fddabe5258d32f90cb3b6adc3df/Makefile#L21-L23).

assets/scss/main.css.scss
```diff
@@ -13,7 +13,11 @@
   
 );
 @forward "uswds";
+@forward "footer";
 
+body {
+  -webkit-font-smoothing: antialiased;
```
aduth, line 19: I think `identity-site` has this style as a remnant of its original implementation, but the design system (both LGDS and USWDS) have moved away from it as a discouraged practice, so I don't know that we should include it here.

Related:

* https://github.com/uswds/uswds/issues/5180
* https://github.com/uswds/uswds/pull/5250
* https://github.com/18F/identity-design-system/pull/362#issuecomment-1699263924

assets/scss/_colors.scss
```diff
@@ -0,0 +1,7 @@
+@use 'uswds-core' as *;
+
+$grey-darker: color('base-darker');
+
+$blue-light: color('primary-light');
```
aduth, line 5: I'd personally favor calling `color` directly rather than creating these variables, since:

1. Variables create a lot of redundancy
2. They're at risk of being incomplete (e.g. [`.footer` styles](https://github.com/GSA-TTS/identity-dev-docs/pull/478/files#diff-dd99657d7065f7185a4acbd6744dc46d4b9d13433c15b566c8e535e8184ff25dR5) call `color('primary-lightest');` directly, so why is that not a variable as well?)
3. Using [tokens](https://designsystem.digital.gov/design-tokens/color/overview/) is more semantic and avoids tying our code to specific colors, e.g. if we were to rebrand, "primary" is still applicable, but "blue" may not be

_includes/footer-navigation.html
Sgtpluck: yes, exactly! and yeah, we can remove `nav` entirely

_includes/footer.html
ajfarkas: Oh wow, these replacements _are_ necessary. `markdownify` adds `p` tags around anything it inserts, so we go an extra level deep, which breaks some CSS (and is bad HTML).

assets/scss/_variables.scss
ajfarkas: File removed. I was only using font-sizing vars, so I moved them to `_footer.scss`.

assets/scss/main.css.scss
ajfarkas: Great. Another reviewer asked me to add in font-smoothing, so it's good to have this documentation.

_data/settings.yml
ajfarkas: (sorry, wrong text box.) I saw this format in `identity-site`, and thought this was a thing we do. Noted.

_data/nav.yml
```diff
@@ -19,3 +19,15 @@ primary:
     href: /security-events/
   - text: Support
     href: /support/
+footer:
```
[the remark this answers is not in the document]
ajfarkas: Removed. I had taken these out of our HTML after doing a cleanup pass of this file.

_includes/footer.html
Sgtpluck: ah, ok, thanks for investigating

assets/scss/_footer.scss
```diff
@@ -0,0 +1,59 @@
+@use 'uswds-core' as *;
+
+$base-font-size: 1rem;
+$base-font-sm: ($base-font-size * 0.8125);
```
aduth, line 4: Minor: Seems like we could use a [`font-body-2xs` design system utility class](https://designsystem.digital.gov/utilities/font-size-and-family/) to get a standard size close to this ad hoc sizing (`.81rem` vs. `.8125rem`). But also I'm not really sure the rationale behind having this small in the first place, and would almost rather just let it stay as normal size. Not like we're running out of space here.

Not a blocker since I assume we're just copying from `identity-site`, but I think the application here is a bit different since we don't have links in this region like we have on the brochure site.

assets/scss/_footer.scss
```diff
@@ -0,0 +1,59 @@
...
+  max-width: 75rem;
+  margin: {
+    left: auto;
+    right: auto
+  }
+  padding: {
+    left: 3.5rem;
+    right: 3.5rem;
+  }
+}
```
aduth, line 20: Is there any reason we can't use a standard [design system grid](https://designsystem.digital.gov/utilities/layout-grid/) here, instead of recreating these styles?

_data/settings.yml
aduth, line 1: I could see these data files being useful if we wanted the site to be localized, but... do we plan to do that? Would it be clearer (less indirection) if these texts were just inlined directly into the markup?

assets/scss/_footer.scss
ajfarkas: I'm happy to have this footer be the baseline. Let me get the ok from Product to change the size, so this doesn't get bounced back to me later.

_data/settings.yml
ajfarkas: I went back and forth on this. I could imagine a scenario where we would have partners who would be more comfortable with the page in Spanish, for instance. Whether we will ever have the budget and will to get translations for dev docs and implement them is another question.

There is some argument that having these text strings as variables in their own file makes it easier to edit and to see any copy errors before publishing, but that's purely nice to have.

aduth: Yeah I think my general feelings on this are a mix of (1) [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and (2) we already have a precedent for embedding content directly in the markup, so having mixed approaches could cause confusion to someone to know where to look for the content.

assets/scss/_footer.scss
ajfarkas: I switched to the design system grid: it looks like we prefer semantic classes over stylesheets, so that's what I went with.  
It's not my personal preference, so if I got that wrong, please let me know, I'm happy to do all styling in SASS.

_includes/footer.html
```diff
@@ -0,0 +1,80 @@
+<footer class="footer grid-container">
```
aduth, line 1: We may need to adjust where exactly the grid is applied in the markup to allow the background color to take effect on the full width. Right now, the footer background doesn't extend the full width on my large monitor:

image

ajfarkas: huh, I may need to requisition a larger monitor.

assets/scss/_footer.scss
```diff
@@ -0,0 +1,82 @@
...
+
+$widescreen: 1400px;
+
+.footer {
+  @include u-padding(0);
+  background-color: color('primary-lightest');
+}
+
+.footer-navigation, .usa-identifier__inner {
+  max-width: $widescreen;
```
aduth, line 11: I would have expected the grid styles would be able to handle this for us.

I was tinkering a bit locally, and this diff seems to work?

```diff
diff --git a/_includes/footer-navigation.html b/_includes/footer-navigation.html
index 9f76150..0616f87 100644
--- a/_includes/footer-navigation.html
+++ b/_includes/footer-navigation.html
@@ -1,2 +1,2 @@
-<div class="footer-navigation">
+<div class="footer-navigation grid-container">
   <div class="footer-navigation__section margin-y-2">
diff --git a/_includes/footer.html b/_includes/footer.html
index 9052c9c..4d7fa34 100644
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -3,3 +3,2 @@
   <div class="usa-identifier">
-    <div class="usa-identifier__inner">
       <section class="usa-identifier__section usa-identifier__section--masthead"
@@ -78,3 +77,2 @@
       </section>
-    </div>
   </div>
diff --git a/assets/scss/_footer.scss b/assets/scss/_footer.scss
index be7199e..7dc8aa4 100644
--- a/assets/scss/_footer.scss
+++ b/assets/scss/_footer.scss
@@ -2,4 +2,2 @@
 
-$widescreen: 1400px;
-
 .footer {
@@ -9,13 +7,2 @@ $widescreen: 1400px;
 
-.footer-navigation, .usa-identifier__inner {
-  max-width: $widescreen;
-
-  @include at-media('widescreen') {
-    margin: {
-      left: auto;
-      right: auto;
-    }
-  }
-}
-
 .footer-navigation {
@@ -56,15 +43,2 @@ $widescreen: 1400px;
 
-.usa-identifier__inner {
-  display: flex;
-  flex-wrap: wrap;
-}
-
-.usa-identifier__section {
-  flex: 1 1 auto;
-}
-
-.usa-identifier__section--masthead, .usa-identifier__section--usagov {
-  flex-grow: 0;
-}
-
 .system-status {
```

aduth (dismissed):

assets/scss/_footer.scss
ajfarkas: Thanks, that's a lot better, I didn't like that `$widescreen` variable in there.
I'm still getting used to the USWDS Grid System.

aduth: >I'm still getting used to the USWDS Grid System.

Me too 😄

aduth (approved):

PR #486, opened 2024-06-03
Make Active Directory a proper noun

n1zyy:
It's not my favorite product in the world, but Active Directory is a proper noun.


n1zyy:
I didn't sign the initial commit so this cannot be merged. I opened #487 to replace this, rather than try to amend and force-push. @zachmargolis, might you do the honors again?

PR #490, opened 2024-06-26
LG-12800 axe a11y fixes

ajfarkas:
## Ticket
[https://cm-jira.usa.gov/browse/LG-12800](https://cm-jira.usa.gov/browse/LG-12800)  
related to: [LGDS MR !2](https://gitlab.login.gov/lg/identity-design-system/-/merge_requests/2)

## Description
Update code to fix all accessibility errors reported by Axe automated test.


ajfarkas:
Ah, this will fail until the LGDS update is made live

aduth:
> Ah, this will fail until the LGDS update is made live

Depending on your timeframe tolerances, you could also temporarily patch in the code to override the design system styles, with the expectation that you'd remove that patch later on once it's published upstream.

I do this pretty frequently in the IdP, for example:

https://github.com/18F/identity-idp/blob/4259cf5f108472cff180d7e8e716ee9a499a0dce/app/components/badge_component.scss#L4-L7

spec/e2e/accessibility.test.js
```diff
@@ -42,22 +43,14 @@ describe('accessibility', () => {
 
   paths.forEach((path) => {
     test(path, async () => {
+      // redirects are causing testing issues
```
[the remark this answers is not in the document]
ajfarkas: Good call, I didn't know we generated that file.
This will be faster than the cudgel I had in place.

ajfarkas:
Bonus: now we're testing the home page `/`, which we weren't before.

Jeremy1026 (approved):

spec/e2e/accessibility.test.js
```diff
@@ -1,18 +1,20 @@
...
-  .map((path) => dirname(relative('_site', path)))
-  .filter((path) => path !== '.')
-  .map((path) => `/${path}/`);
+// only test canonical paths
+const paths = (() => {
+  const file = readFileSync('./_site/sitemap.xml', { encoding: 'utf8' });
+  const pathMatch = file.matchAll(/[.gov|:4000](\/[A-Za-z-]+\/?[A-Za-z-]*\/)<\/loc/g);
+  const foundPaths = Array.from(pathMatch).map((a) => a[1]);
+  foundPaths.unshift('/');
+  return foundPaths;
```
aduth, line 16: Rather than hardcoding the port here (which can change with local developer configuration or CLI flags), maybe we parse it as a URL and use the pathname? Also avoids having to manually add in `/`.

```suggestion
  const pathMatch = file.matchAll(/<loc>(.+?)<\/loc>/g);
  return Array.from(pathMatch).map(([, url]) => new URL(url).pathname));
```

assets/scss/main.css.scss
```diff
@@ -146,6 +148,11 @@ button.code-button:hover {
   width: 230px;
 }
 
+.usa-nav__close .usa-icon {
```
aduth, line 151: What was this needed for?

assets/scss/main.css.scss
```diff
@@ -15,6 +15,8 @@
 @forward "uswds";
 @forward "footer";
 
+$block-background-color: #fafafa;
```
aduth, line 18: I see this is copied from the design system styles, but I might plan to surface this as it's strange we don't use one of our standard design system tokens (like a `base-lightest` or something).

_plugins/content_code.rb
```diff
@@ -0,0 +1,24 @@
+require 'rouge'
+
+# Include tabindex for accessibility reasons
+class Rouge::Formatters::HTMLPygmentsA11y < Rouge::Formatters::HTMLPygments
+  def stream(tokens, &b)
+    yield %(<div class="highlight"><pre class="#{@css_class}"><code tabindex="0">)
```
aduth, line 6: As discussed when we paired, I personally think it'd be better to put the tabindex on the element with the `overflow` styling, which also helps avoid some strange focus outline wrapping visuals.

```suggestion
    yield %(<div class="highlight"><pre class="#{@css_class}" tabindex="0"><code>)
```

_plugins/content_code.rb
```diff
@@ -0,0 +1,24 @@
+require 'rouge'
+
+# Include tabindex for accessibility reasons
```
aduth, line 3: Might be helpful to point to the original implementation for future maintainer reference.

```suggestion
# Include tabindex for accessibility reasons
# See: https://github.com/rouge-ruby/rouge/blob/master/lib/rouge/formatters/html_pygments.rb
```

_pages/user-experience/sign-in-sign-out.md
```diff
@@ -94,7 +94,7 @@ Do use the white button with an outline on light backgrounds that don’t provid
 </ul>
 
 <div class="lookbook-embed-container margin-top-4">
-<iframe class="lookbook-embed" src="https://idp.dev.identitysandbox.gov/components/embed/login_button/workbench?_options=%257B%2522scenarios%2522%253A%255B%2522workbench%2522%255D%252C%2522panels%2522%253A%255B%2522params%2522%252C%2522output%2522%255D%257D"></iframe>
+<iframe title="Login.gov button lookbook" class="lookbook-embed" src="https://idp.dev.identitysandbox.gov/components/embed/login_button/workbench?_options=%257B%2522scenarios%2522%253A%255B%2522workbench%2522%255D%252C%2522panels%2522%253A%255B%2522params%2522%252C%2522output%2522%255D%257D"></iframe>
```
aduth, line 97: "lookbook" is the tool we use, but doubtful it'd be easily understood by an end-user.

```suggestion
<iframe title="Login.gov button preview" class="lookbook-embed" src="https://idp.dev.identitysandbox.gov/components/embed/login_button/workbench?_options=%257B%2522scenarios%2522%253A%255B%2522workbench%2522%255D%252C%2522panels%2522%253A%255B%2522params%2522%252C%2522output%2522%255D%257D"></iframe>
```

aduth (approved):
Mostly minor comments, LGTM overall!

assets/scss/main.css.scss
ajfarkas: This was part of the update to LGDS 9.3.0

aduth: I wouldn't expect you'd need to implement any custom styles beyond what's already included in the design system. Were these missing from the design system  styles?

_pages/user-experience/sign-in-sign-out.md
ajfarkas: Lookbook is a common term. I understood it without knowing it was a tool.
Preview isn't exactly right, either, since there's more detail. I think this is one of those infamous naming problems.

README.md
```diff
@@ -28,9 +28,9 @@ Run the tests:
 make test
 ```
 
-This site uses the [U.S. Web Design Standards](https://standards.usa.gov). To update them:
+This site uses the [U.S. Web Design System](https://designsystem.digital.gov/). To update them:
```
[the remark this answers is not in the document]
ajfarkas: I looked into this: we do still use USWDS, and the process, as outlined here, is to manually add it to the project.
My concern if we remove this is that people will forget to do that.
This seems like tech debt we should address.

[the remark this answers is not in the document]
ajfarkas: scratch that, we don't strictly need that.

spec/e2e/accessibility.test.js
ajfarkas: hm, this catches `/risk.json`, which is not an HTML file and causes the tests to fail. I like this direction, I'll dig into it more.

assets/scss/main.css.scss
ajfarkas: Sorry, I was trying to get that comment in before a meeting.
LGDS 9 removed the close SVG for `` tags, and replaced it with `svg > icon`. It's a drop-in replacement, but we had a custom size on the `img`, and so I'm adding that here, to match what's on `main`.

_plugins/content_code.rb
ajfarkas: I can resolve this with `code { display: inline-block; }`. This lets me keep the focus on the proper element.

aduth: Can you elaborate on "proper element"? The focus should be on the scrollable element, which is the `<pre>`.

ajfarkas:
@aduth I'm working on these in branch `LG-12800_small_updates`

_plugins/content_code.rb
ajfarkas: Is the `pre` the scrollable element because it has set dimensions?  
I tested this, and you are correct: the `pre` element is the one that is scrolling. That's not intuitive to me, because the `code` element is the one with content. I guess if it had a text node I wouldn't think that way.

I'll go ahead and make the change you suggested.

aduth: It's because [the `pre` is the element that has the `overflow` styling](https://github.com/18F/identity-design-system/blob/main/src/scss/packages/usa-code/src/_styles.scss#L35-L37), where the overflow styling will cause scroll if needed ([ref](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x#auto)).

ajfarkas:
PR for small updates at: https://github.com/GSA-TTS/identity-dev-docs/compare/main...LG-12800_small_updates

ajfarkas:
PR with the small changes is [#491](https://github.com/GSA-TTS/identity-dev-docs/pull/491)

PR #493, opened 2024-08-08
Add more detail to the JWT description

nprimak:
A partner contacted us wanting clarification about the JWT in the token endpoint and how we validate it. This drew attention to the fact that our docs don't specify this info (that the private key should be associated with the public key uploaded to the dashboard) 

Ticket in Zendesk: https://logingov.zendesk.com/agent/tickets/9715

### DIFF
--- a/_pages/oidc/token.md
+++ b/_pages/oidc/token.md
@@ -23,7 +23,7 @@ sidenav:
 
 ---
 {% capture client_assertion %}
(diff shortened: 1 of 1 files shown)


ajfarkas (commented):
This is a very complex topic for a lot of partners: we should bring in someone with less technical experience here, and get suggestions for improving this.

karenstrother:
> This is a very complex topic for a lot of partners: we should bring in someone with less technical experience here, and get suggestions for improving this.

Hey! I'm here!

karenstrother (commented):
As the non-technical expert this is very clear. It is plain language and explains the connection between these parts.

zachmargolis (approved):

PR #506, opened 2024-10-30
[LG-14462] Add footer links

h-m-m:
## Changes

Added footer links to match what exists on the main site. I tried to copy and paste as much as applicable from https://github.com/GSA-TTS/identity-site and then reconcile any apparent CSS duplication. I did not try to include the translation-based options from identity-site as that seemed like scope creep for this ticket.

I added the font anti-aliasing setting that's on the main page with the intention of keeping the sites looking consistent.

## Screenshots

<details><summary>The previous dev docs footer</summary>

Screenshot 2024-10-30 at 17 48 57

</details>

<details><summary>The new dev docs footer</summary>

Screenshot 2024-10-30 at 17 47 35

</details>

<details><summary>The current Login.gov footer</summary>

Screenshot 2024-10-30 at 17 46 33

</details>

You may notice that the left and right margins / padding on the main site footer are different than those on the dev docs. This was true before this PR. Fixing them in this PR seemed to me like it would likely be more complicated than the rest of the effort on this PR so far a relatively small benefit, so I felt stopping here was good enough for now.

<details><summary>Before font anti-aliasing</summary>

Screenshot 2024-10-30 at 17 55 07

</details>
<details><summary>After font anti-aliasing</summary>

Screenshot 2024-10-30 at 17 55 24

</details>

Especially in screenshots, I find the font anti-aliasing more noticeable in the body paragraph text than in the footer itself, so here's a comparison. The anti-aliased version is definitely closer visually to what's on the main site.

### DIFF
--- a/_includes/footer-navigation.html
+++ b/_includes/footer-navigation.html
@@ -13,7 +13,7 @@
       aria-label="Footer navigation"
       class="grid-row grid-gap-2 tablet:grid-gap-6"
     >
-      {% assign login_base = 'https://login.gov' %}
+      {% assign login_base = 'https://www.login.gov' %}
       {% for section in site.data.nav.footer_sections %}
         <section class="grid-col-6 tablet:grid-col-auto margin-y-2 footer-navigation__section">
           <h2>{{ section.heading }}</h2>
--- a/_includes/footer-navigation.html
+++ b/_includes/footer-navigation.html
@@ -1,5 +1,5 @@
-<div class="footer-navigation grid-container">
-  <div class="footer-navigation__section margin-y-2">
+<div class="footer-navigation grid-container grid-row">
+  <div class="footer-navigation__section grid-col-12 margin-y-2 tablet:grid-col-auto">
     <img
       src="{{ site.baseurl }}/assets/img/logo.svg"
       alt="Login.gov logo"
@@ -7,7 +7,28 @@
       height="27"
     />
   </div>
-  <div class="footer-navigation__section">
+  <div class="grid-col-fill"></div>
+  <div class="grid-col-12 tablet:grid-col-auto">
+    <nav
+      aria-label="Footer navigation"
+      class="grid-row grid-gap-2 tablet:grid-gap-6"
+    >
+      {% assign login_base = 'https://login.gov' %}
+      {% for section in site.data.nav.footer_sections %}
+        <section class="grid-col-6 tablet:grid-col-auto margin-y-2 footer-navigation__section">
+          <h2>{{ section.heading }}</h2>
+          <ul>
+            {% for section_link in section.links %}
+              <li>
+                <a href="{{ section_link.url | replace: '$BASE_URL', login_base }}">
+                  {{ section_link.label }}
+                </a>
+              </li>
+            {% endfor %}
+          </ul>
+        </section>
+      {% endfor %}
+    </nav>
     <div class="system-status">
       <a class="font-body-2xs" href="https://status.login.gov">Login.gov system status</a>
     </div>
--- a/assets/scss/_footer.scss
+++ b/assets/scss/_footer.scss
@@ -6,14 +6,9 @@
 }
 
 .footer-navigation {
-  display: flex;
   justify-content: space-between;
-  flex-wrap: wrap;
-  @include u-padding-top(3);
-  @include u-padding-right(2);
-  @include u-padding-bottom(3);
-  @include u-padding-left(2);
-
+  @include u-padding-top(2);
+  @include u-padding-bottom(5);
   @include at-media('desktop') {
     @include u-padding-right(4);
     @include u-padding-left(4);
@@ -23,18 +18,20 @@
     @include unstyled-list;
   }
 
+  h2,
   a {
+    font-size: 13px;
     color: color('primary-dark');
   }
 }
 
-.footer-navigation__section {
-  flex: 0 0 100%;
+.footer-navigation__section h2 {
+  margin: 0;
+}
 
-  @include at-media('tablet') {
-    display: flex;
-    flex-basis: auto;
-  }
+.footer-navigation__section a {
+  @include u-margin-top(2);
+  display: block;
 }
 
 .usa-identifier__usagov-description {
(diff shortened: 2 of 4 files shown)


ajfarkas:
I recall some discussion around the width of the footer. We have some different styles in dev docs that made the wider footer make sense…

ajfarkas (approved):
great, lgtm

assets/scss/_footer.scss
```diff
@@ -23,26 +18,32 @@
     @include unstyled-list;
   }
 
+  h2,
   a {
+    font-size: 13px;
```
ajfarkas, line 23: I missed this, I think our font-sizes are supposed to be in rem, so they resize with user default changes.

ajfarkas (changes requested):

assets/scss/main.css.scss
```diff
@@ -18,6 +18,10 @@
 
 $block-background-color: #fafafa;
 
+body {
+  -webkit-font-smoothing: antialiased;
+}
+
```
aduth, line 24: I would suggest not including this. Font smoothing has been removed from both USWDS and the Login.gov Design System intentionally.

For history, see:

* https://github.com/18F/identity-design-system/pull/362#issuecomment-1699263924
* https://github.com/uswds/uswds/pull/5250

```suggestion
```

h-m-m: Thanks for this history, @aduth, I found this super helpful.

I pulled this in because it's currently active in https://github.com/GSA-TTS/identity-site/blob/a2a42a1d00d1d58dc5c45c5efaa4cbf7aba00898/_sass/components/_typography.scss#L8

I personally think the text is slightly easier to read without it, but I assumed consistency was more important.

assets/scss/_footer.scss
h-m-m: Thanks for catching this. I must have copied this from the browser inspector's calculated value instead of the supplied value in https://github.com/GSA-TTS/identity-site/

h-m-m: Here's the source: 

https://github.com/GSA-TTS/identity-site/blob/a2a42a1d00d1d58dc5c45c5efaa4cbf7aba00898/_sass/_variables.scss#L4-L5

assets/scss/main.css.scss
aduth: Yeah, it probably ought to be removed from the brochure site as well.

assets/scss/_footer.scss
h-m-m: I was under the impression that there's a potential problem with rem and font sizes where, since rem are relative, certain contexts can cause the same rem font sizes to be different actual sizes of font within the same page. That said

* we clearly use it regularly in other code bases
* I like consistency
* I would not be surprised if our CSS standard practices are such that the potential problem is guaranteed never to arise

I'll change this to 0.85rem, since that seems to be the closest multiple of 0.05rem

h-m-m:
Screenshots after latest updates


<details><summary>Mobile breakpoint</summary>

Screenshot 2024-11-01 at 12 20 30

</details>

<details><summary>Middle breakpoint</summary>

Screenshot 2024-11-01 at 12 18 23


</details>


<details><summary>Above 1024px</summary>

Screenshot 2024-11-01 at 12 18 35


</details>

assets/scss/_footer.scss
ajfarkas: if you use `em`, you risk having 1em being different px sizes on the same page, because that is relative to the "m" character of the font you're currently using at the current font size. So, if I have a `div` set to 16px, then a `p` set to 1.5em, that's 24px; if that `p` has a `span` that's 1.5em, the size of an "m" on the parent is 24px, so the span text is 36px.

`rem` is "root em", which is relative to the font-size of the root, in the case of an HTML document, the `html` element. Because that is static throughout the document, 1.5em is always the same px. The trick here is that users can adjust the root size in their browser settings (or with extensions), which then causes every element's font-size to scale the same amount, which makes it very useful for allowing users to "zoom in" the text without having to scroll sideways as they would if they zoomed the whole page.

ajfarkas: the `ex` unit is relative to the height of the "x" character, which is not often used in Web development but when you need it, is a fantastic trick.

h-m-m: Ah! So in concern about this potential problem, I was getting `em` and `rem` switched. Got it. Unsure where I got the idea that it was "relative em" instead of "root em"

nprimak (approved):

Gemfile.lock
```diff
@@ -122,7 +122,7 @@ DEPENDENCIES
   rspec_junit_formatter
 
 RUBY VERSION
-   ruby 3.2.5p208
+   ruby 3.2.6p234
```
nprimak, line 125: Did you update this because of the CircleCI Ruby version error? I merged in a PR on Friday that fixes that so you might not need this anymore.

ajfarkas (approved):

_includes/footer-navigation.html
```diff
@@ -1,13 +1,34 @@
...
     />
   </div>
-  <div class="footer-navigation__section">
+  <div class="grid-col-fill"></div>
+  <div class="grid-col-12 tablet:grid-col-auto">
+    <nav
+      aria-label="Footer navigation"
+      class="grid-row grid-gap-2 tablet:grid-gap-6"
+    >
+      {% assign login_base = 'https://login.gov' %}
```
aduth, line 16: Minor: While technically both versions work, the [configured canonical domain should be `www.`](https://github.com/GSA-TTS/identity-site/blob/a2a42a1d00d1d58dc5c45c5efaa4cbf7aba00898/_config.yml#L4), and ideally we'd have redirects set up to force this, where we'd want to avoid an unnecessary redirect hop ([LG-10513](https://cm-jira.usa.gov/browse/LG-10513)). 

```suggestion
      {% assign login_base = 'https://www.login.gov' %}
```

Gemfile.lock
h-m-m: Yes. Though since this works, I don't see a compelling reason to roll back. Happy to revert if you think I should

h-m-m: Rebased just to make sure nothing breaks and I do have to roll this back after your Friday changes

nprimak: Thanks! Sorry for the extra trouble.

PR #510, opened 2024-11-06
[LG 14519] Change Dev Docs language from Dashboard to Portal

rahul-kamarouthu:
[LG 14519] Change Dev Docs language from Dashboard to Portal

Jira ticket - https://cm-jira.usa.gov/browse/LG-14519

### DIFF
--- a/_data/errors.yml
+++ b/_data/errors.yml
@@ -8,7 +8,7 @@ oidc:
         The issuer (also known as the client id) provided does not match a service provider registered in the IdP.
         ##### What to do:
         - Ensure the authentication request has a `client_id` value
-        - Ensure that the `client_id` sent in the authentication request matches the one registered in your application on the [Partner Dashboard](https://dashboard.int.identitysandbox.gov/){:target="_blank"}.
+        - Ensure that the `client_id` sent in the authentication request matches the one registered in your application on the [Partner Portal](https://dashboard.int.identitysandbox.gov/){:target="_blank"}.
         - Ensure that the application configuration's `active` value is set to `true`.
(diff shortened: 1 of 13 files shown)


ajfarkas (changes requested):
One instance of "Dashboard" in `_pages/overview.md`. I'm not entirely sure that the text is correct, since the link is to IDP and will not redirect the partner to the Portal.

But this does raise the question of whether you should update the URL from `dashboard.int.identitysandbox.gov` to `portal.int.identitysandbox.gov`, especially where that address is displayed in text (on `_pages/testing.md`)

Screenshot 2024-11-07 at 3 39 33 PM

rahul-kamarouthu:
> One instance of "Dashboard" in `_pages/overview.md`. I'm not entirely sure that the text is correct, since the link is to IDP and will not redirect the partner to the Portal.
> 
> But this does raise the question of whether you should update the URL from `dashboard.int.identitysandbox.gov` to `portal.int.identitysandbox.gov`, especially where that address is displayed in text (on `_pages/testing.md`)
> 
> Screenshot 2024-11-07 at 3 39 33 PM

Changed!

ajfarkas:
The issue remains that the last link you updated is incorrect: it does not take the user to the Partner Portal.

I also think it's worth checking whether we should be updating the URL here, so that users are not hovering over "Partner Portal" and seeing `https://dashboard.int.identitysandbox.gov`.

ajfarkas (approved):
Let's get this moving.
Please make sure we have a ticket to update the remaining URL's to the `portal` subdomain.

PR #522, opened 2024-11-20
Fix payload structure of account disabled event

mitchellhenke:
The `reason` key is outside of the `subject`, and the example payload is correct, but the documentation currently has it inside. This PR also removes the extra period.

IDP for reference:
https://github.com/18F/identity-idp/blob/74f0ff081fb96eb02bb4f2665a01c7648f114f11/app/services/push_notification/account_disabled_event.rb#L18-L27
| Before                       | After                                                                         |
|-----------------------------|-------------------------------------------------------------------------------------|
| image | image |


Jeremy1026 (approved):

ajfarkas (approved):
lgtm

PR #526, opened 2024-12-07
Update token.md

manpreetbhasin:


mitchellhenke:
Thanks for opening a PR!

In this context, I do think client can be kind of ambiguous and we could consider updating the language to be clearer. However, the token exchange does not necessarily happen on the client-side from the perspective of the relying party.

Open to discussion, but without further explanation on the intended clarification, I'm going to close the PR for now.

PR #556, opened 2025-07-03
use template for remaining accordions

mmagsa:
Part 2 of the accordion refactor started in [555](https://github.com/GSA-TTS/identity-dev-docs/pull/555), addressing the remaining accordions.
Also fixes a couple of small bugs (add missing `</div>` and `&`->`&amp;` fix)


h-m-m:
Do you mind including screenshots of this in the PR? Even after-the-change screenshots by themselves are nice. I think I've said this before, but I personally find them very helpful when trying to figure out what someone intended with a past code change when next year browsers or other dependencies may change so the page no longer renders like it used to.

Jeremy1026 (approved):

PR #557, opened 2025-07-08
add links to accordions

mmagsa:
Ticket: [[identity-dev-docs] Add ability to easily copy link to accordion](https://gitlab.login.gov/lg-teams/FIE/team-fie-daily/-/issues/17)

This change adds a little link icon with a copyable link (visible only when hovering over the accordion) to the bottom left corner. Pasting this link into the browser's address bar takes you to the accordion on corresponding page and opens it.

Before
image

After
image

### DIFF
--- a/_includes/accordion.html
+++ b/_includes/accordion.html
@@ -14,7 +14,7 @@ include
       {{ include.title }}
       <a href="#{{ include.id }}"
          class="accordion-link"
-         aria-label="copy"
+         aria-label="anchor"
          aria-describedby="{{ include.id }}"
          >#</a>
     </button>


Jeremy1026 (approved):

ajfarkas:
#

**🔨 issue:** ‏The alt-text on all of the anchor links is the same, "anchor".  
The alt-text should be descriptive, and/or next to the title of the section. This is a rough experience for programmatic users.

ajfarkas:
> > Suggestion: Should this issue be fixed in a separate bug ticket then, instead of partially in this? Can you elaborate on the fragility you're seeing that adding links to the template would improve?
> 
> Since this issue seems to have been an issue prior to this ticket work I think it should be a follow-up ticket to fix the anchors throughout the dev docs to be A11y.

I think the anchor links have several issues, including the location. It makes more sense to update these links to be correct, and then update the existing problematic links in a separate ticket, rather than to approve a ticket with known tech-debt.

Sgtpluck (changes requested):
i would like to see the anchor tags coded in html, not javascript, for the reasons that i enumerated in my [comment](https://github.com/GSA-TTS/identity-dev-docs/pull/557/files#r2195364190). 

i'd also like the placement of the anchor tag to be towards the top of the content, which is where i would intuitively expect it to be. while this might be geared towards our onboading engineers, it's still a website that our partners use, and so a link anchor should be towards the top of the content for both visual consistency and accessibility.

ajfarkas:
> On the `aria-label="anchor"` I've found [this discussion](https://github.com/bryanbraun/anchorjs/issues/63) which seems to indicate that this is ok. Also, since most of the accordions are always closed, and the link is currently inside the body of the accordions, shouldn't screen readers not pick those up? 🤔

I didn't get the same conclusion from that discussion. In formal testing, it's explicitly non-accessible to have many links with the same text. [The ICT Baseline describing the rule](https://ictbaseline.access-board.gov/web-baselines/14Links/#accessibility-requirements) actually uses anchorjs as well, but programmatically describes the link with the heading text.

We work really hard to make the Portal accessible without JS, it doesn't make sense to rely on JS for the Dev Docs to be readable, especially given that it is a static site. If we are relying on JS today, our goal should be to remove that dependency, and all changes moving forward should only use JS as a progressive enhancement.

_includes/accordion.html
```diff
@@ -12,6 +12,7 @@
       aria-controls="{{ include.accordion_id }}"
     >
       {{ include.title }}
+      <a href="#{{ include.accordion_id }}" class="accordion-link">#</a>
```
ajfarkas, line 15: What the ICT Baseline page does is to include `aria-label="copy"` and `aria-describedby={usa-accordion__button.id}`
This way it shows as "copy {FAQ title}" in the Link rotor, and reads as "copy" (without duplicating the FAQ title) in the flow of the page.

assets/js/anchor_accordion.js
```diff
@@ -1,16 +1,17 @@
 (() => {
-  const onChange = (_event) => {
+  const jumpToAccordion = (_event) => {
     const urlHash = new URL(document.URL).hash;
-    const idFromHash = urlHash.match(/^(#[A-Za-z0-9_-]+)$/);
+    const idFromHash = urlHash.match(/^#([A-Za-z0-9_-]+)$/);
```
ajfarkas, line 4: **💭 thought:** ‏This could be streamlined, if you want

1. `urlHash.substr(1)`
2.  `urlHash.replace('#','')`
These just make it so you don't have to be thorough with your regex.  
‏Maybe easier is to use `accordionBody = document.querySelector(urlHash)`. The performance hit is not measurable for a single execution.

assets/js/anchor_accordion.js
```diff
@@ -1,16 +1,17 @@
...
     if (idFromHash) {
-      const accordionHeading = document.querySelector(
-        `${idFromHash[1]}.usa-accordion__heading > button`,
-      );
-      if (accordionHeading) {
-        accordionHeading.click();
+      const accordionBody = document.getElementById(idFromHash[1]);
+      const accordionHeading = accordionBody.previousElementSibling;
+      const accordionButton = accordionHeading.querySelector('.usa-accordion__heading > button');
```
ajfarkas, line 8: **💭 thought:**  This could be simplified by removing `.usa-accordion__heading >` since you're querying the selector from within that element already.‏

assets/js/anchor_accordion.js
```diff
@@ -1,16 +1,17 @@
...
+      if (accordionButton) {
+        accordionButton.click();
       }
     }
   };
-  document.addEventListener('DOMContentLoaded', onChange);
-  window.addEventListener('hashchange', onChange);
+
+  document.addEventListener('DOMContentLoaded', jumpToAccordion);
+  window.addEventListener('hashchange', jumpToAccordion);
```
ajfarkas, line 16: **🔨 issue:** This `hashchange` listener breaks the opening functionality.

Clicking on an anchor link refreshes the page, so that is covered by both `DOMContentLoaded` and  `hashchange`, which opens and then closes the accordion.  
`hashchange` also works when you manually update the hash in the URL bar (which does not reload the DOM). 

In order to catch all of these cases, you can simplify the `if (idFromHash)` block to:
```js
const accordionBody = document.getElementById(idFromHash[1]);
accordionBody.hidden = false;
```

spec/integration/accordion_js_spec.rb
```diff
@@ -11,15 +11,15 @@
   end
 
   it 'is expanded with an anchor in the URL' do
```
ajfarkas, line 13: **🎯 suggestion:** ‏There is another case to capture, which is updating the URL hash only. I'm not sure how rspec handles that, but it's clear that `visit` refreshes the page (where updating only the hash would not).

ajfarkas:
> Also - any thoughts on the `#` instead of the little chainlink icon (which is supplied by anchorjs, so I didn't feel like I should spend time on figuring out how to use it here)?
The octothorpe is great. No sense in spending time to add the link icon.

Sgtpluck (approved):
thanks for making the change to html! i think that's nicer. we should probably audit the use of anchorjs at all, but that's out of scope of this ticket.

as for the icon vs `#` -- maybe we should be using the uswds link icon [for both](https://designsystem.digital.gov/components/icon/)? but if we want to keep em simple i'd be more in favor updating the anchor js icon to be the [octothorp](https://www.bryanbraun.com/anchorjs/#octothorp) rather than trying to get the chainlink icon.

_includes/accordion.html
ajfarkas: I think you could go back to `anchor`, as you have the added context where needed.

assets/js/anchor_accordion.js
ajfarkas: You can do a check `if (accordionBody.hidden) accordionButton.click();`.  
The issue is that you have to account for the state, because that `click` handler is a toggle: if the accordion is already open (as it is in 1 of 3 cases), it will close it, which is not what you want.

The other thing it does here is to update `aria-expanded` on the associated button. You can see it in the dev tools.

PR #583, opened 2025-12-15
Dev-Docs #11 Capitalize Partner Portal, lowercase portal

ajfarkas:
[Dev-Docs #11](https://gitlab.login.gov/lg-teams/FIE/developer-docs/-/issues/11)

Make sure that all instances of "Partner Portal" are capitalized, and other general instances of "portal" are lowercase.


_includes/support/faq_csp_violation.html
```diff
@@ -4,9 +4,9 @@ <h5>Background:</h5>
...
     <br/><br/>
-    This error occurs when Service Providers attempt to redirect users to a url that is not registered in the Redirect URLs field in the <a target="_blank" href="https://dashboard.int.identitysandbox.gov/">Login.gov Portal</a> application's configuration. All urls that users could be redirected to, even as a passthrough, need to be included in the list of Redirect URLs.
+    This error occurs when Service Providers attempt to redirect users to a url that is not registered in the Redirect URLs field in the <a target="_blank" href="https://dashboard.int.identitysandbox.gov/">Login.gov portal</a> application's configuration. All urls that users could be redirected to, even as a passthrough, need to be included in the list of Redirect URLs.
```
mmagsa: Question: outside of the scope of this ticket, but "Login.gov portal" seems... wrong, and it looks like we keep using that in various places. Some of these instances should probably be renamed to "Partner Portal", and others may need to be rephrased. @nickttng thoughts?

ajfarkas: This is pretty small, there are 6 instances across 3 pages. Should we go with Partner Portal, since the original is somewhat formal, as "Login.gov portal"?

ajfarkas: I will say there is at least one spot where "Login.gov Partner Portal" makes sense because it is in context with other portals (ie, 'MS Power Apps Portal')

PR #599, opened 2026-03-05
Fix example URL in risc_outgoing.yml

davidbsanderson:
The reproof-completed event was referencing the wrong URL in the example.


mmagsa (approved):
Thanks for catching and fixing!

PR #606, opened 2026-03-31
Update RISC API docs

monfresh:
- Add the `exp` claim to the example payload
- Remove the Identifer Recycled and Identifer Changed events as we plan on
getting rid of them. See this Slack thread:
https://gsa-tts.slack.com/archives/CA7NE63SB/p1747664974521479

### DIFF
--- a/_data/risc_outgoing.yml
+++ b/_data/risc_outgoing.yml
@@ -102,50 +102,6 @@
     }
   }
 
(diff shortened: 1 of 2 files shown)


nickttng (approved):
From my end, looks good to me

mmagsa (approved):
Looks good to me!

PR #609, opened 2026-04-28
chore: use single quotes consistently in individual state error YAML

jgravois:
hey y'all! 👋 

i ran into a couple unexpected errors today proofing a mock identity in the sandbox. i'm not positive that the double-quotes in your sample snippet were causing the problem but i figured it'd be nice to use single quotes consistently like you do in other similar files.


mmagsa (approved):
Looks good to me. Thank you for helping out!

PR #637, opened 2026-08-31
Consolidate release notes headers

nprimak:
I spoke with Nick about how to handle the headers in the Release Notes page and he suggested we could just have 1 header since they were a little redundant anyway.


ajfarkas (approved):
Looks good!

PR #640, opened 2026-09-11
Add Chinese (Simplified) as locale option in OIDC authorization requests

mitchelled:
Hello! 🙂

I was reviewing the documentation while evaluating a potential integration, and noticed a small potential oversight related to the `locale` OIDC authorization request parameter. My understanding is Chinese (Simplified) is also supported, and I tested to confirm that it indeed is.

Going from 2 to 3 in the list feels unwieldy in the current structure, but I didn't want to propose more changes than necessary.


_pages/oidc/authorization.md
```diff
@@ -200,7 +200,7 @@ In an **unsuccessful authorization**, the URI will contain the parameters `error
           <h4 class="clearfix">locale</h4><span class="float-left text-italic">optional</span>
         </div>
         <div class="grid-col-7">
-          <p>If you know that a user would prefer one of our alternative language translations (currently Spanish or French), you can include the <code class="language-plaintext highlighter-rouge">locale</code> parameter to specify the language Login.gov should use (either <code class="text-lowercase">ES</code> for Spanish or <code class="text-lowercase">FR</code> for French).</p>
+          <p>If you know that a user would prefer one of our alternative language translations (currently Spanish, Chinese (Simplified), or French), you can include the <code class="language-plaintext highlighter-rouge">locale</code> parameter to specify the language Login.gov should use (<code class="text-lowercase">ES</code> for Spanish, <code class="text-lowercase">ZH</code> for Chinese, or <code class="text-lowercase">FR</code> for French).</p>
```
jules-fsa: Wonder if it should now be updated as `If you know that a user would prefer one of our alternative language translations — currently Spanish, Chinese (Simplified), or French —`

mitchelled: Thanks, I made some edits to do that as well as some other small changes.

